PR #27, opened 2016-07-21
gradlew assembleDebug throws autolink library not found - fixed

tridib:
org.nibor.autolink library was missing. Build was failing.


tridib:
Posted an issue to see if others are facing same issue.

PR #41, opened 2016-11-04
Introduce checkstyle/pmd/findbugs

YusukeIwaki:
#40 for `develop2` branch (special thanks to @laggedHero !)

PR #42, opened 2016-11-04
Fix style and rules

YusukeIwaki:
base: #41

**Rules updated:** https://github.com/RocketChat/Rocket.Chat.Android/pull/42/commits/4e101ae7fdfe7ccd3e6efcc739c2513cfe8b0f7a

### `CustomImportOrder`
  REMOVED because it is too severe. (Android Studio plugin doesn't help re-ordering import as Google-Java-style says)

### `allowedAbbreviationLength `, `packageName`
  temporary-REMOVED for current speedy development.

### DIFF
--- a/config/quality/pmd/pmd-ruleset.xml
+++ b/config/quality/pmd/pmd-ruleset.xml
@@ -20,8 +20,8 @@
     <rule ref="rulesets/java/naming.xml">
         <!--<exclude name="AbstractNaming" />-->
         <exclude name="LongVariable" />
-        <exclude name="ShortMethodName" />
-        <exclude name="ShortVariable" />
+        <!--exclude name="ShortMethodName" /-->
+        <!--exclude name="ShortVariable" /-->
         <!--<exclude name="ShortClassName" />-->
         <!--<exclude name="VariableNamingConventions" />-->
     </rule>
--- a/config/quality/pmd/pmd-ruleset.xml
+++ b/config/quality/pmd/pmd-ruleset.xml
@@ -4,25 +4,25 @@
     xsi:noNamespaceSchemaLocation="http://pmd.sf.net/ruleset_xml_schema.xsd"
     xsi:schemaLocation="http://pmd.sf.net/ruleset/1.0.0 http://pmd.sf.net/ruleset_xml_schema.xsd">
 
-    <description>Custom ruleset for Rocket.Chat.Android application</description>
+  <description>Custom ruleset for Rocket.Chat.Android application</description>
 
-    <exclude-pattern>.*/R.java</exclude-pattern>
-    <exclude-pattern>.*/gen/.*</exclude-pattern>
+  <exclude-pattern>.*/R.java</exclude-pattern>
+  <exclude-pattern>.*/gen/.*</exclude-pattern>
 
-    <rule ref="rulesets/java/android.xml" />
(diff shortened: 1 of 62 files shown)


laggedHero:
@YusukeIwaki That's no fix. The AOSP project does not follows the [Google Java Style](https://google.github.io/styleguide/javaguide.html).

Since we're not writing against AOSP we don't need (and should not) follow its Java style.

YusukeIwaki:
@laggedHero Thank you for your advices.

It is right that AOSP style is quite different from Google Java Style.
However I think some rules (such as `MemberName` or `StaticVariableName`) are more useful than Google's style, especially for Android developers.

For example, considering the lifecycles of Activity/Fragment or something, it would be good practice to indicate member valiables by staring the name with "m" to distinguish it from local variables.

How do you think?

config/quality/pmd/pmd-ruleset.xml
```diff
@@ -20,8 +20,8 @@
     <rule ref="rulesets/java/naming.xml">
         <!--<exclude name="AbstractNaming" />-->
         <exclude name="LongVariable" />
-        <!--<exclude name="ShortMethodName" />-->
-        <!--<exclude name="ShortVariable" />-->
+        <exclude name="ShortMethodName" />
```
laggedHero: I leave those two lines commented out because short naming should be an exception. Things named like `s` or `c` are not good for reading. Even in a couple of lines.

laggedHero:
About Hungarian Notation: no, I don't see it as a useful thing when using moderns IDE (like AS). They make a really great work in showing the scope. Not only highlighting with different colors but allowing to quick inspect by holding `Cmd` (Mac, sorry) and hover over it with the mouse.

config/quality/pmd/pmd-ruleset.xml
YusukeIwaki: Yes, I agree `s` or `c` is not proper name, however how about `id`?

```
    public static void showServerConfigActivity(Context context, String id) {
```

With the ShortVariable rule, the code above would raise error. But I thought

```
    public static void showServerConfigActivity(Context context, String serverConfigId) {
```

is a little redundant.

laggedHero: I like the "redundant" version. It makes the method signature more explicit about what we're passing to it. A nice bonus, if you ask me.

As a last resort, we could simply [suppress the warning](http://pmd.sourceforge.net/snapshot/usage/suppressing.html).

YusukeIwaki: I got it. I will revert the modification 😆 for pmd rule.

Maybe some Realm model classes will have `@SuppressWarning("PMD.ShortVariable")` because DDP often uses short JSON keys like `id` or `t` and they are mapped to Realm object as it is.

YusukeIwaki:
@laggedHero I just understood that some Hungarian Notation used in AOSP is not usually used in Android application development, by browsing some famous apps or developers.
Thank you for your kind advice. I will fix codes.

laggedHero:
Jake Wharton has a nice post/rant about it. I always laugh when I read it. xD

YusukeIwaki:
@laggedHero I replaced checkstyle config with google-java-style based one.

However I think some rules are too severe at this time, and currently they are excluded.
Could you check them?

PR #46, opened 2016-11-23
Stops the same object from been propagated over and over as if updated

laggedHero:
@RocketChat/android

I noticed that the user image would blink over and over again. So I checked and saw that the `isSame` would always return `false`.

I changed thinks a little bit in order to use the Realm provided `equals`.

### DIFF
--- a/realm-helpers/src/main/java/chat/rocket/android/realm_helper/RealmObjectObserver.java
+++ b/realm-helpers/src/main/java/chat/rocket/android/realm_helper/RealmObjectObserver.java
@@ -22,7 +22,7 @@ public class RealmObjectObserver<T extends RealmObject> extends AbstractRealmRes
     }
 
     protected boolean isSame(T element1, T element2) {
-      return false;
+      return element1.equals(element2);
     }
   }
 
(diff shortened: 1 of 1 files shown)


realm-helpers/src/main/java/chat/rocket/android/realm_helper/RealmObjectObserver.java
```diff
@@ -22,7 +22,7 @@ protected T extractObjectFromResults(RealmResults<T> results) {
     }
 
     protected boolean isSame(T element1, T element2) {
-      return false;
+      return element1.equals(element2);
```
YusukeIwaki, line 25: I'm not sure, but I suspect `element1.equals(element2)` is always false.
I will check if so.

realm-helpers/src/main/java/chat/rocket/android/realm_helper/RealmObjectObserver.java
```diff
@@ -47,29 +47,31 @@ public void setImpl(Impl<T> impl) {
...
-        T target = source != null ? realm.copyFromRealm(source) : null;
-        if (previousResult != null && impl.isSame(previousResult, target)) {
+      @Override
+      public void onChange(RealmResults<T> element) {
+        T currentResult = impl.extractObjectFromResults(element);
+        if (previousResult != null && impl.isSame(previousResult, currentResult)) {
           return;
         }
-        previousResult = target;
+        previousResult = currentResult;
```
YusukeIwaki, line 64: I suspect it is dangerous to preserve the Realm object without detaching from DB...

YusukeIwaki (commented):
I will check your code, thank you!

realm-helpers/src/main/java/chat/rocket/android/realm_helper/RealmObjectObserver.java
YusukeIwaki: > I changed thinks a little bit in order to use the Realm provided equals.

Right!
I checked `element1.equals(element2)` is true.

YusukeIwaki:
```
11-25 14:44:03.467 E/AndroidRuntime( 8902): java.lang.IllegalStateException: Object is no longer valid to operate on. Was it deleted by another thread?
11-25 14:44:03.467 E/AndroidRuntime( 8902): 	at io.realm.internal.UncheckedRow.nativeGetIndex(Native Method)
11-25 14:44:03.467 E/AndroidRuntime( 8902): 	at io.realm.internal.UncheckedRow.getIndex(UncheckedRow.java:123)
11-25 14:44:03.467 E/AndroidRuntime( 8902): 	at io.realm.RoomSubscriptionRealmProxy.equals(RoomSubscriptionRealmProxy.java:981)
11-25 14:44:03.467 E/AndroidRuntime( 8902): 	at chat.rocket.android.realm_helper.RealmObjectObserver$Impl.isSame(RealmObjectObserver.java:25)
```

it seems to be required for object to detach from DB ...

PR #51, opened 2016-11-25
FIX #50 fix the behavior of back-press on SlidingPane is opened.

### LINKED ISSUE #50: [side menu active] pressing the back button (physical one) should close it
it should close the side menu if it's active, instead of loading the last page. if the side menu is not active, it should do the normal behavior

YusukeIwaki:
@RocketChat/android

Closes #50

The logic for handling Back-key in MainActivity was:

1.  If fragment handles it, consumed.
2.  otherwise; close Activity.

Now it will be:

1.  **close SlidingPane if possble with consume.**
2.  If not, if fragment handles it, consume.
3.  otherwise; close Activity.


laggedHero:
Sounds good. Let's merge it!

PR #52, opened 2016-11-25
Changed the lifrcycle events used for the sidebar subscriptions

laggedHero:
@RocketChat/android

Trying to better understand what would trigger the update in the sidebar I found out that we're using `onPause` and `onResume`. I think that `onViewCreated` and `onDestroyView` makes more sense here.

### DIFF
--- a/app/src/main/java/chat/rocket/android/fragment/sidebar/SidebarMainFragment.java
+++ b/app/src/main/java/chat/rocket/android/fragment/sidebar/SidebarMainFragment.java
@@ -2,6 +2,7 @@ package chat.rocket.android.fragment.sidebar;
 
 import android.os.Bundle;
 import android.support.annotation.Nullable;
+import android.view.View;
 import android.widget.CompoundButton;
 import android.widget.ImageView;
 import android.widget.LinearLayout;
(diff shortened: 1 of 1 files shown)


YusukeIwaki:
Thank you for your suggestion.
I agree with your change if GarbageCollection doesn't run.

But actually, realm instance often and suddenly dies due to GC... 
So it is not a good idea to keep Realm observers alive in background.

Or do you have any idea to prevent GC working and keep Realm alive...? 😅

laggedHero:
If I got it right: I have a fragment, not GC'e, with a reference to realm. Realm is GC'e even so?

YusukeIwaki:
Yes, even if the realm is referred by Activity/Fragment, the realm instance is suddenly CLOSED when GC runs. And, Realm observer often stop working suddenly...

laggedHero:
I'll deny this since it looks to be troublesome. I really need to better understand realm. Thx

YusukeIwaki:
@laggedHero Sorry, I found why Realm instance is closed suddenly.

`Realm#isClosed()` is not always true even after `Realm#close()` called.
So RealmObjectObserver and other Realm helper components sometimes illegally closed the Realm instance duplicately.

This pull request will be acceptable after I fix RealmHelper, RealmObjectObserver, and RealmListObserver!

PR #53, opened 2016-11-25
Exported strings

laggedHero:
@RocketChat/android

Clean up the view a bit too.

In order to support i18n and so on we need to export all the strings to the strings.xml


YusukeIwaki:
Great! 👍 
Thank you for your contribution!

PR #88, opened 2016-12-16
Feature/unread mode

### LINKED ISSUE #81: [v1.0-beta] Unread rooms - required for usable UX
Your Rocket.Chat.Android version: latest alpha

The Unread rooms: feature is required to be usable.   For any users with a moderate amount of activities on a server similar to demo.rocket.chat  (almost all real users).

This is **not obvious** in testing envirnoments, but in production ALL of us have 100's of channels that we monitor.  Without bubbling of the channels that has new unread messages, it will be impossible to be productive with the app.

screen shot 2016-12-09 at 1 02 53 pm

laggedHero:
@RocketChat/android

Closes #81 

While this do shows the unread mode, it dos not changes it dynamically. I'm listening to changes on the user object and that won't work quite the way we need here.

We'll probably need to rework things around here.

### DIFF
--- a/app/src/main/java/chat/rocket/android/layouthelper/chatroom/RoomListManager.java
+++ b/app/src/main/java/chat/rocket/android/layouthelper/chatroom/RoomListManager.java
@@ -3,61 +3,60 @@ package chat.rocket.android.layouthelper.chatroom;
 import android.view.View;
 import android.view.ViewGroup;
 
-import java.util.List;
 import chat.rocket.android.helper.TextUtils;
 import chat.rocket.android.model.ddp.RoomSubscription;
 import chat.rocket.android.widget.internal.RoomListItemView;
+import java.util.List;
 
 /**
  * Utility class for mapping Room list into channel list ViewGroup.
  */
 public class RoomListManager {
+  private View unreadTitle;
+  private ViewGroup unreadRoomsContainer;
   private ViewGroup channelsContainer;
   private ViewGroup dmContainer;
+
+  private boolean unreadRoomMode = false;
+
+  private List<RoomSubscription> roomSubscriptionList;
+
+  /**
+   * Callback interface for List item clicked.
+   */
+  public interface OnItemClickListener {
+    void onItemClick(RoomListItemView roomListItemView);
+  }
+
   private OnItemClickListener listener;
+
+  /**
+   * constructor with three ViewGroups.
+   */
+  public RoomListManager(View unreadTitle, ViewGroup unreadRoomsContainer,
+                         ViewGroup channelsContainer, ViewGroup dmContainer) {
+    this(unreadTitle, unreadRoomsContainer, channelsContainer, dmContainer, false);
+  }
 
   /**
    * constructor with two ViewGroups.
    */
-  public RoomListManager(ViewGroup channelsContainer, ViewGroup dmContainer) {
+  public RoomListManager(View unreadTitle, ViewGroup unreadRoomsContainer,
+                         ViewGroup channelsContainer, ViewGroup dmContainer,
+                         boolean unreadRoomMode) {
+    this.unreadTitle = unreadTitle;
+    this.unreadRoomsContainer = unreadRoomsContainer;
     this.channelsContainer = channelsContainer;
     this.dmContainer = dmContainer;
-  }
-
-  private static void removeItemIfExists(ViewGroup parent, String roomName) {
-    for (int i = 0; i < parent.getChildCount(); i++) {
-      RoomListItemView roomListItemView = (RoomListItemView) parent.getChildAt(i);
-      if (roomName.equals(roomListItemView.getRoomName())) {
-        parent.removeViewAt(i);
-        break;
-      }
-    }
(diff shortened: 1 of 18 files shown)


YusukeIwaki:
It is not good to rename member names of RealmObject.

Object extending RealmObject should have the same structure as the JSON in DDP payload.
If not, Realm cannot automatically parse and map JSON into the object using `realm#createOrUpdateFromJson` or `realm#createOrUpdateAllFromJson`.

YusukeIwaki (changes requested):
It is not good to rename member names of RealmObject.

Object extending RealmObject should have the same structure as the JSON in DDP payload.
If not, Realm cannot automatically parse and map JSON into the object using realm#createOrUpdateFromJson or realm#createOrUpdateAllFromJson.

laggedHero:
I don't think this is a problem. Realm is using reflection and the object field name. So we may use the getters and setters do better use.

Have you tried to pull this and run locally? Did it broke something?

YusukeIwaki:
@laggedHero Sorry for my misunderstanding. I will check again.

Realm used to have severe restriction for the naming of getters/setters, however the restriction no longer exists after 0.88. I didn't know the fact. Sorry...

dependencies.gradle
```diff
@@ -5,8 +5,7 @@ ext {
   retroLambdaPatch = 'me.tatarka.retrolambda.projectlombok:lombok.ast:0.2.3.a2'
   compileSdkVersion = 25
   buildToolsVersion = '25.0.1'
-  minSdkVersionDebug = 21
-  minSdkVersion = 17
+  minSdkVersion = 16
```
YusukeIwaki, line 8: API 16 doesn't support RTL features (such as `marginStart`, `marginEnd`, and so on)
For supporting API 16, we should replace them with app-compat expressions (such as `android:layout_marginStart` to `app:layout_marginStart` )

app/src/main/java/chat/rocket/android/model/ddp/User.java
```diff
@@ -22,12 +22,13 @@
   private String status;
   private double utcOffset;
   private RealmList<Email> emails;
+  private Settings settings;
```
YusukeIwaki, line 25: I think it is better to have `settings` with String type (just JSON string).

Realm automatically creates a new record of settings and preference infinitely for each updating the user logged in.

image

For avoiding this, we have to set `@PrimaryKey` into settings and preference. But it is impossible...

So storing settings as raw JSON string seems to be an only way to prevent increasing records.

app/src/main/java/chat/rocket/android/layouthelper/chatroom/RoomListManager.java
```diff
@@ -126,6 +133,40 @@ private void onItemClick(View view) {
...
+
+    for (RoomSubscription roomSubscription : roomSubscriptionList) {
+      String name = roomSubscription.getName();
+      if (TextUtils.isEmpty(name)) {
+        continue;
+      }
+
+      String type = roomSubscription.getType();
+
+      if (unreadRoomMode && roomSubscription.getUnread() > 0) {
```
YusukeIwaki: `roomSubscription.isAlert()` would be better.

slice

https://github.com/RocketChat/Rocket.Chat/blob/d70c4e2/packages/rocketchat-ui-sidenav/side-nav/unreadRooms.coffee#L10

YusukeIwaki:
This pull request includes some changes not related to implement unread feature.

However they are really important for readable code and acceptable enough. 👍 
So I cherry-picked them in advance.

Could you please rebase this pull request?

app/src/main/java/chat/rocket/android/layouthelper/chatroom/RoomListManager.java
geekgonecrazy: Downside to that is isAlert only surfaces the ones we are mentioned in right?  If unread mode for sure want to show them all.  Now if we put mentioned at the very top... That wouldn't hurt anyone's feelings I don't think

YusukeIwaki (approved):

dependencies.gradle
laggedHero: Good call!

I'll add it as a task to do here.

Looks like a small price to pay to be able to reach 4.5% (https://developer.android.com/about/dashboards/index.html - as of Dec 19) of the devices out there. :)

laggedHero:
@YusukeIwaki please... don't cherry pick stuff like that again. It's giving me a headache here.  xD

YusukeIwaki:
@laggedHero very sorry for inconvenience. I will not do like this again.

app/src/main/java/chat/rocket/android/model/ddp/User.java
laggedHero: I have a solution for this. I'll update the branch with the code.

app/src/main/java/chat/rocket/android/layouthelper/chatroom/RoomListManager.java
laggedHero: If the web app is using `isAlert()` we should do the same. I'll update it here.

PR #89, opened 2016-12-16
Added calls for setting the away presence

### LINKED ISSUE #87: Perma online status
Your Rocket.Chat.Android version: Latest

I've been scratching my head trying to figure this one out.

Steps to reproduce:

* Sign out of all sessions.  (Killed on desktop / all mobile devices) Logged in with another user and confirmed it does show I am signed out.
* Login using the Mobile app
* Close the Mobile app.  I swiped it away.  I noticed immediately it shows i'm offline.  But then a few seconds later it shows that i'm online again.

If I then Force Quit I go completely offline.

Currently using the other app for push notifications, so when this shows me as connected Rocket.Chat doesn't send push notifications.  I'm not sure if its opening a new connection as a service in the background?

Oddly looking at app info it says 0% of battery used by this.

laggedHero:
@RocketChat/android

Closes #87

### DIFF
--- a/app/src/main/java/chat/rocket/android/api/MethodCallHelper.java
+++ b/app/src/main/java/chat/rocket/android/api/MethodCallHelper.java
@@ -278,6 +278,15 @@ public class MethodCallHelper {
         .onSuccessTask(task -> Task.forResult(null));
   }
 
+  public Task<Void> setUserAway() {
+    return call("UserPresence:away", TIMEOUT_MS, JSONArray::new)
+        .onSuccessTask(task -> Task.forResult(null));
+  }
+
+  public Task<Void> setUserOnline() {
+    return call("UserPresence:online", TIMEOUT_MS, JSONArray::new)
+        .onSuccessTask(task -> Task.forResult(null));
+  }
 
   public Task<JSONObject> getUsersOfRoom(final String roomId, final boolean showAll) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/chat/rocket/android/api/MethodCallHelper.java
```diff
@@ -278,6 +278,15 @@ public MethodCallHelper(RealmHelper realmHelper, DDPClientWraper ddpClient) {
         .onSuccessTask(task -> Task.forResult(null));
   }
 
+  public Task<Void> setUserAway() {
+    return call("UserPresence:away", TIMEOUT_MS, JSONArray::new)
```
YusukeIwaki, line 282: `MethodCallHelper#call(name, timeout)` is also available here. And it would better to use it.

YusukeIwaki (approved):

app/src/main/java/chat/rocket/android/api/MethodCallHelper.java
YusukeIwaki: I will fix this after merging!

laggedHero: No, please.. don't "fix" it.

I'll grab my computer on Monday and then we talk about it.

PR #94, opened 2016-12-20
render Markdown link

### LINKED ISSUE #83: [v1.0-beta] render link, image-link
render

`[link](http://demo.rocket.chat)`

`image`

ref: RocketChat: packages/rocketchat-markdown/markdowncode.coffee

YusukeIwaki:
@RocketChat/android

Closes #83

* hide `image ` because it is rendered as attachment.
* render links with `[link](http://example.com/)` and `<http://example.com/|link>` formatted only not in single-line code.

slice

### DIFF
--- a/rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/helper/InlineHightlighter.java
+++ b/rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/helper/InlineHightlighter.java
@@ -13,6 +13,8 @@ import android.text.style.StyleSpan;
 import android.text.style.TypefaceSpan;
 import android.widget.TextView;
 
+import java.util.regex.Matcher;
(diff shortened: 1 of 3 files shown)


YusukeIwaki:
`<http://example.com/|link>` causes crash. I will fix it.

geekgonecrazy:
Yeah I don't think the web client supports that.  All those links we get from slack in imports we transform them

YusukeIwaki:
slice

fixed.

laggedHero:
@YusukeIwaki What about caching those compiled patterns? =

We could make each pattern a static field.

What do you think?

YusukeIwaki:
@laggedHero Thank you for good advice.

It should be better to define Pattern as static. I will fix it.

laggedHero (approved):

PR #99, opened 2016-12-21
Changed message entry field

### LINKED ISSUE #93: [v1.0-beta] Redesign message entry
Our message entry field should be always available and not a fab enabled button. (And we should not have more than one fab like button on screen).

We should try to do something like the web app (see attached image bellow)

screen shot 2016-12-19 at 5 52 03 pm

(ps.: @engelgabriel asked for this. I think it makes sense.)

laggedHero:
@RocketChat/android

Please review the guidelines for contributing to this repository.
Closes #93 

That's how it looks now.

screen shot 2016-12-21 at 5 49 51 pm

screen shot 2016-12-21 at 5 49 59 pm

screen shot 2016-12-21 at 5 50 06 pm


YusukeIwaki:
Awesome! 👍  I will check the code :)

YusukeIwaki:
Great! 👍 

However it is a little hard to understand the responsibility of the classes with the name `MessageSelection...`  or `...MessageSpec...`.
I will refactor the naming of some classes after merging this.

YusukeIwaki (approved):
👍

laggedHero:
@YusukeIwaki yep they are not good, but not that bad.

We'll need to define a way to plugin more message types as needed. So my initial idea was a message specification (`MessageSpec`). There would be defined a way to enter the message (pick a file), send it to the server and render it back.

In the end, the task was taking too long and I just pushed it like this.

PR #102, opened 2016-12-22
fix issue that login screen is not shown.

YusukeIwaki:
@RocketChat/android

Login screen is shown only when MainActivity is foreground.

```
    sessionObserver = realmHelper
        .createObjectObserver(realm ->
            Session.queryDefaultSession(realm)
                .isNotNull("token")
                .equalTo("tokenVerified", true)
                .isNull("error"))
        .setOnUpdateListener(session -> {
          if (session == null && isForeground) {
            LaunchUtil.showServerConfigActivity(this, serverConfigId);
          }
        });
```

This is triggered on
`MainActivity#updateSessionObserver`
 triggered by `MainActivity#onServerConfigIdUpdated`
 triggered by `AbstractAuthedActivity#onResume`

So isForeground should be turned true **before** `AbstractAuthedActivity#onResume` !!


app/src/main/java/chat/rocket/android/activity/MainActivity.java
```diff
@@ -53,14 +53,14 @@ protected void onStart() {
 
   @Override
   protected void onResume() {
-    super.onResume();
     isForeground = true;
+    super.onResume();
   }
 
   @Override
   protected void onPause() {
```
laggedHero, line 61: I think we may leave the `isForeground = false;` before the call to super here. Looks like a bit more safe.

YusukeIwaki: Exactly. I will fix it. Thank you.

laggedHero (approved):

PR #109, opened 2016-12-22
Refactor message composer

YusukeIwaki:
@RocketChat/android

slice

### design change

* upload action picker dialog is moved into rocket-chat-android-widgets, and callbacks are simplified.
* AbstractMessageSpec just implements 2 interfaces. one is for callback (MessageExtraActionBehavior), another is for view presenter (MessageExtraActionItemPresenter).

### rename

* MessageComposer -> MessageFormLayout
* MessageSpec, MessageSelection -> MessageExtraAction


rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/message/MessageFormLayout.java
```diff
@@ -5,43 +5,51 @@
...
 import android.widget.TextView;
 
+import bolts.Continuation;
+import bolts.Task;
 import chat.rocket.android.widget.R;
+import chat.rocket.android.widget.internal.ExtraActionSelectionDialog;
 
-public class MessageComposer extends LinearLayout {
+public class MessageFormLayout extends LinearLayout {
+  public static final int EXTRA_ACTION_PICK_PHOTO = 1;
```
laggedHero: The message type cannot be static. It needs to be dynamic.

We need to define a plugin-like arch to support it.

laggedHero (changes requested):

rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/message/MessageFormLayout.java
YusukeIwaki: Thank you for good advice.

I failed to consider that each server has different configurations about uploading feature.
I will fix...

laggedHero (approved):

PR #110, opened 2016-12-22
Checks for the server version

### LINKED ISSUE #90: [v1.0-beta] Verify server version and notify user of compability
We should check the server version and notify the user if the server is not compatible with the app.

Maybe we may include some link to docs... ?

laggedHero:
@RocketChat/android

Closes #90 

Pops up a little toast saying the server is invalid.

### DIFF
--- a/app/src/main/java/chat/rocket/android/fragment/server_config/InputHostnameFragment.java
+++ b/app/src/main/java/chat/rocket/android/fragment/server_config/InputHostnameFragment.java
@@ -9,7 +9,7 @@ import chat.rocket.android.R;
 import chat.rocket.android.RocketChatCache;
 import chat.rocket.android.helper.LogcatIfError;
 import chat.rocket.android.helper.OkHttpHelper;
-import chat.rocket.android.helper.ServerHelper;
+import chat.rocket.android.helper.ServerPolicyHelper;
 import chat.rocket.android.helper.TextUtils;
 import chat.rocket.android.model.ServerConfig;
 import chat.rocket.android.realm_helper.RealmObjectObserver;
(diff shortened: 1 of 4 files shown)


app/src/main/java/chat/rocket/android/helper/ServerHelper.java
```diff
@@ -0,0 +1,103 @@
...
+import org.json.JSONObject;
+
+import java.io.IOException;
+import okhttp3.Call;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+import okhttp3.ResponseBody;
+
+public class ServerHelper {
```
YusukeIwaki: The name of `ServerHelper` is vague.
`ServerVersionChecker` or another name (with the class responsibility) would be better.

laggedHero: Indeed.. `Helper` says nothing at all. Good thing we do have PR's now to get our attention at that.

Thx

YusukeIwaki (approved):

app/src/main/java/chat/rocket/android/fragment/server_config/InputHostnameFragment.java
```diff
@@ -37,11 +40,42 @@ protected void onSetupView() {
...
+          public void isNotValid() {
+            getActivity().runOnUiThread(() ->
+                Toast.makeText(getActivity(), R.string.input_hostname_invalid_server_message,
+                    Toast.LENGTH_SHORT).show());
+          }
+        });
+  }
+
+  @Override
+  public void onResume() {
```
YusukeIwaki, line 62: onResume should be removed if only calling the method of super class.

laggedHero: I didn't even see that. Was not in the part of the code I was changing.

We should probably do it the next time we touch that class. :)

PR #120, opened 2016-12-28
	From string to constants

laggedHero:
@RocketChat/android

Adding some constants and using it.


YusukeIwaki:
@laggedHero Great! 

But could you provide "push" feature with another pull request?
Replacing strings to constants is very easy to accept, however "push" feature requires some discussion.

YusukeIwaki:
https://github.com/cmelchior/realmfieldnameshelper

Does this library help defining constants?

laggedHero:
I don't think that library will help. Some of our fields are just a letter (`t` as an example). Let's create those constants.

About `push`: looks like some of my initial work leaked here. My bad. But those are not a lot of stuff. Any problems you see here?

YusukeIwaki:
> I don't think that library will help. Some of our fields are just a letter (t as an example). Let's create those constants.

I see 👍 

> About push: looks like some of my initial work leaked here. My bad. But those are not a lot of stuff. Any problems you see here?

"Push" feature is very important for us.
It should be better not to mix incomplete codes with other pull requests even if the code is easy to understand.

So I created a commit for removing "push" features from this pull request.
https://github.com/RocketChat/Rocket.Chat.Android/pull/120/commits/c389c82db95cd2d0bfd1b8c24866aeb5a35779ff

After merging this pull request, the commit can be reverted with other "push" related codes.

YusukeIwaki:
Actually I prefer constants to be defined in another (or inner) class

```
class Message extends RealmObject {
  private String rid;
  private String msg;

  public interface ColumnNames {
    String ROOM_ID = "roomId";
    String MESSAGE = "msg";
  }
}
```

or with prefix to indicate that it is column name.

```
class Message extends RealmObject {
  private String rid;
  private String msg;

  public static final String COLUMN_ROOM_ID = "roomId";
  public static final String COLUMN_MESSAGE = "msg";
}
```

because `Message.MESSAGE` or `Message.USER` is sometimes hard to guess it have column name without checking the definition of Message class.

YusukeIwaki (approved):

Issue #77, opened 2016-12-09
Push notifications technically impossible?


kramer65:
Since push notifications with FCM are not supported in the Cordova Android app (and new GCM tokens are not distributed anymore) I was stoked to see the development of a native Android app. So I was following the development closely, when I just found this little note at the [end of the feature list](https://github.com/RocketChat/Rocket.Chat.Android/blob/doc/feature.md#wont-be-supported):

> ## won't be supported
>  - push notification via FCM (technically impossible)
> 

I'm really puzzled by this. Next to the fact that push notifications are the main reason we're waiting for this native app, I am also wondering; why are push notifications for this app specifically technically impossible. Seeing that there are about a million apps out there supporting push notifications, I wouldn't know why this would be technically impossible for this specific app.

Any info on this would be very welcome!

YusukeIwaki:
It is possible to implement notification using continuous WebSocket connection and will be supported in v1.0-beta (#72)

But we cannot implement it using FCM.

For implementing it, we have to build app with **one** FCM sender id. And we can't configure multiple FCM sender ids for one app.
But each Rocket.Chat server has different FCM sender id...

So we can't implement app with FCM push notification that can be used for arbitrary servers.

Sing-Li:
@YusukeIwaki    we understand that there maybe problem in supporting notification Gateway with the native app (although admittedly difficult to understand why since the Cordova app already does it)

@engelgabriel   but we DO need to be able to use FCM for any community-user-private build (or white label) of the native app  ... that *is* the official Android means of mobile notification intake

Also, please test the battery drain characteristics of a "continuous websocket notification scheme" before committing to develop it.    Battery drain is the NUMBER ONE reason why users perfer native app over Cordova.    (many can put up with the slightly laggish performance)

geekgonecrazy:
@YusukeIwaki The no multiple FCM sender id is for sure a problem.  To get around this with our cordova apps on Android and iOS we had to use a gateway.  All Rocket.Chat Servers out of the box point to the gateway server.  This allows people to right off the bat use our Official mobile apps.

So every rocket.chat server sends to our gateway.  The Cordova app has our GCM sender id compiled in.  The gateway relays to GCM and then it delivers.  This prevents any need for us to introduce any sort of continuous connection.

graywolf336:
Yes, I feel we should continue with this model for the native applications because that's the number one reason people use mobile applications.

laggedHero:
[...]

PR #124, opened 2017-01-02
Push notifications using the Cordova plugin code

laggedHero:
Trying to keep up with the same behavior with the existing Cordova app

@RocketChat/android

First step for  #77 (still need to make FCM work)

I've tried to separate the message receiving (in this case, from GCM) from the processing and displaying of the notification. Should make it easier to integrate with FCM later (minus, maybe, the already set expected params for notification).

Notification actions are still a incognita as of now.


app/src/main/java/chat/rocket/android/activity/ServerConfigActivity.java
```diff
@@ -77,14 +83,27 @@ protected void onPause() {
...
+        .continueWith(task -> {
+          registerForPush();
+          return task;
+        })
+        .continueWith(task -> {
+          onRenderServerConfigSession(session);
+          return task;
+        })
+        .continueWith(new LogcatIfError());
+  }
```
YusukeIwaki: Registering push seems not to be dependent on Activity.
So it is better to have the logic in `SessionObserver#onLogin`

app/src/main/java/chat/rocket/android/service/RocketChatWebSocketThread.java
```diff
@@ -52,7 +51,6 @@
       NewMessageObserver.class,
       CurrentUserObserver.class,
       ReactiveNotificationManager.class,
-      NotificationItemObserver.class,
```
YusukeIwaki, line 55: Why is this removed?

app/src/main/java/chat/rocket/android/push/gcm/RegistrationIntentService.java
```diff
@@ -0,0 +1,81 @@
...
+import chat.rocket.android.RocketChatCache;
+import chat.rocket.android.api.MethodCallHelper;
+import chat.rocket.android.model.ServerConfig;
+import chat.rocket.android.model.ddp.PublicSetting;
+import chat.rocket.android.model.ddp.PublicSettingsConstants;
+import chat.rocket.android.model.ddp.User;
+import chat.rocket.android.realm_helper.RealmHelper;
+import chat.rocket.android.realm_helper.RealmStore;
+
+public class RegistrationIntentService extends IntentService {
```
YusukeIwaki: It would be better to rename the class to `GcmRegistrationIntentService`

app/src/main/java/chat/rocket/android/push/gcm/PushInstanceIDListenerService.java
```diff
@@ -0,0 +1,62 @@
...
+    return RealmStore.getDefault().executeTransactionForReadResults(
+        realm -> realm.where(ServerConfig.class).findAll());
+  }
+
+  private void flagForRefresh(List<ServerConfig> serverConfigs) {
+    final RealmHelper realmHelper = RealmStore.getDefault();
+
+    for (final ServerConfig serverConfig : serverConfigs) {
+      final RealmHelper serverRealmHelper = RealmStore
+          .getOrCreate(serverConfig.getServerConfigId());
```
YusukeIwaki: `getOrCreate` is really required?

It seems unnecessary to create if DB not exists.

app/src/main/java/chat/rocket/android/push/gcm/PushInstanceIDListenerService.java
```diff
@@ -0,0 +1,62 @@
...
+    Intent intent = new Intent(this, RegistrationIntentService.class);
+    startService(intent);
+  }
+
+  private List<ServerConfig> getServerConfigs() {
+    return RealmStore.getDefault().executeTransactionForReadResults(
+        realm -> realm.where(ServerConfig.class).findAll());
+  }
+
+  private void flagForRefresh(List<ServerConfig> serverConfigs) {
```
YusukeIwaki: The responsibility of this method cannot be guessed by the name `flagForRefresh`

`updateSyncPushToken(List<ServerConfig>)` or similar one would be better.

app/src/main/java/chat/rocket/android/push/gcm/PushInstanceIDListenerService.java
```diff
@@ -0,0 +1,62 @@
...
+    for (final ServerConfig serverConfig : serverConfigs) {
+      final RealmHelper serverRealmHelper = RealmStore
+          .getOrCreate(serverConfig.getServerConfigId());
+
+      boolean isPushEnable = PublicSetting
+          .getBoolean(serverRealmHelper, PublicSettingsConstants.Push.ENABLE, false);
+      String senderId = PublicSetting
+          .getString(serverRealmHelper, PublicSettingsConstants.Push.GCM_PROJECT_NUMBER, "").trim();
+
+      serverConfig.setSyncPushToken(isPushEnable && !"".equals(senderId));
```
YusukeIwaki: It is better to update syncPushToken in a RealmTransaction.

app/src/main/java/chat/rocket/android/push/PushConstants.java
```diff
@@ -0,0 +1,73 @@
+package chat.rocket.android.push;
+
+public interface PushConstants {
+  String COM_ADOBE_PHONEGAP_PUSH = "com.adobe.phonegap.push";
+  String REGISTRATION_ID = "registrationId";
+  String FOREGROUND = "foreground";
+  String TITLE = "title";
+  String NOT_ID = "notId";
```
YusukeIwaki, line 8: `NOTIFICATION_ID` is better.

app/src/main/java/chat/rocket/android/push/PushNotificationHandler.java
```diff
@@ -0,0 +1,654 @@
...
+
+    NotificationCompat.Builder notificationBuilder = new NotificationCompat.Builder(context)
+        .setWhen(System.currentTimeMillis())
+        .setContentTitle(fromHtml(extras.getString(TITLE)))
+        .setTicker(fromHtml(extras.getString(TITLE)))
+        .setContentIntent(contentIntent)
+        .setAutoCancel(true);
+
+    SharedPreferences prefs = context
+        .getSharedPreferences(PushConstants.COM_ADOBE_PHONEGAP_PUSH, Context.MODE_PRIVATE);
```
YusukeIwaki, line 125: Which component does set these prefs?

app/src/main/java/chat/rocket/android/model/ServerConfig.java
```diff
@@ -31,6 +32,7 @@
   private int state;
   private String session;
   private String error;
+  private boolean syncPushToken;
```
YusukeIwaki, line 35: ServerConfig is originally designed to represent(sync) the connectivity of each server.

Is `syncPushToken` related with the connectivity? If not, this property should be in another RealmObject-based class.

app/src/main/java/chat/rocket/android/api/MethodCallHelper.java
```diff
@@ -304,6 +307,19 @@ public MethodCallHelper(RealmHelper realmHelper, DDPClientWrapper ddpClient) {
...
+      JSONObject param = new PushUpdate(pushId, token, userId).toJson();
+      return new JSONArray().put(param);
+    }).onSuccessTask(task -> Task.forResult(null));
+  }
+
+  public Task<Void> pushSetUser(String pushId) {
+    return call("raix:push-setuser", TIMEOUT_MS, () -> new JSONArray().put(pushId))
+        .onSuccessTask(task -> Task.forResult(null));
+  }
+
```
YusukeIwaki: It is better to define these method in subclass of MethodCallHelper as FileUploadingHelper does.

app/src/main/java/chat/rocket/android/helper/ServerPolicyHelper.java
```diff
@@ -71,6 +71,14 @@ private static String removeProtocol(String hostname) {
     return hostname.replace("http://", "").replace("https://", "");
   }
 
+  private static String removeTrailingSlash(String hostname) {
+    if (hostname.charAt(hostname.length() - 1) != '/') {
```
YusukeIwaki, line 75: `hostname.replaceAll("/+$", "")` removes the slashes at the tail of hostname.
Is this condition really needed?

app/src/main/java/chat/rocket/android/model/params/PushUpdate.java
```diff
@@ -0,0 +1,34 @@
+package chat.rocket.android.model.params;
+
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class PushUpdate {
```
YusukeIwaki: This class is not related to model and only used to the method `MethodCallHelper#pushUpdate`. 

It is better to define this logic as a private method in [the subclass of MethodCallHelper](https://github.com/RocketChat/Rocket.Chat.Android/pull/124/files#r94359793).

YusukeIwaki:
I tried this with [the APK built with CircleCI](https://circleci.com/gh/RocketChat/Rocket.Chat.Android/367#artifacts/containers/0), but couldn't receive push notification by the mention `@YusukeIwaki test` by another user...

app/src/main/java/chat/rocket/android/activity/ServerConfigActivity.java
laggedHero: Do the `SessionObserver#onLogin` runs before the user logs in? If it does, that looks fine.

If not, we could try something else. The idea is to run those calls before the user login.

app/src/main/java/chat/rocket/android/service/RocketChatWebSocketThread.java
laggedHero: Some of the guys were getting upset by this notification. I'll try a build without it and get some feedback.

app/src/main/java/chat/rocket/android/push/gcm/RegistrationIntentService.java
laggedHero: I'll do the same with the instance Id listener.

app/src/main/java/chat/rocket/android/push/gcm/PushInstanceIDListenerService.java
laggedHero: `RealmStore::get` tries to get a cached `RealmConfiguration`. If it's not cached, it will return null.

I'm thinking when the user has more than one server config and we need to refresh the token. I just simply trust `RealmStore::get` for that case.

Later on, when we do have multiple servers running and we get around this issue someother way we may chance this.

app/src/main/java/chat/rocket/android/push/gcm/PushInstanceIDListenerService.java
laggedHero: Since that object is a copy from realm there's no problem doing this and calling `realmHelper.executeTransaction(realm -> realm.copyToRealmOrUpdate(serverConfig));` (called on the next call)

app/src/main/java/chat/rocket/android/push/PushConstants.java
laggedHero: While I did add to the constants, I did not chance the existing ones. That's a class from the Cordova plugin and I would like to keep changes at a minimum.

I'll move my changes to the end fo the file and flag as so.

app/src/main/java/chat/rocket/android/push/PushNotificationHandler.java
laggedHero: The Cordova plugin uses this for some of its configuration. Since we'll have some white label work I thought that it would be nice to leave it there.

laggedHero:
I didn't add CIrcleCI's certs to the Google console... ooops xD

PS.: I'm missing some pushes too. The demo server is busy with something else.

app/src/main/java/chat/rocket/android/helper/ServerPolicyHelper.java
laggedHero: Why running a regex when there's no need to?

app/src/main/java/chat/rocket/android/model/ServerConfig.java
laggedHero: I was trying to leave server related config vars on the `ServerConfig` class.

Maybe I could move it somewhere else and rename this class to `ServerConnectivityConfig`??

app/src/main/java/chat/rocket/android/helper/ServerPolicyHelper.java
YusukeIwaki: The intent (improving performance) is a little hard to guess without any comments, because`removeTrailingSlash` is called not so frequently enough to consider performance improvement.

Please add comment for `if (hostname.charAt(hostname.length() - 1) != '/')`.

app/src/main/java/chat/rocket/android/activity/ServerConfigActivity.java
```diff
@@ -77,14 +83,27 @@ protected void onPause() {
   }
 
   private void onRenderServerConfigError(ServerConfig config) {
-    if (config != null) {
+    if (config.getState() == ServerConfig.STATE_CONNECTION_ERROR) {
```
YusukeIwaki: Why the logic of 'state = STATE_CONNECTION_ERROR' is moved here?
By moving the logic, sessionObserver.sub() is called repeatedly.

app/src/main/java/chat/rocket/android/activity/ServerConfigActivity.java
YusukeIwaki: As `continueWithServerConfigSession` and `SessionObserver#onLogin` uses the same observer implementation, they will be called almost the same moments.

However not guaranteed to be called before user login (even if the logic is put in the Activity)
It depends on the implementation of the server.

```
Thu Dec 08 20:05:25 JST 2016	Thu Dec 08 20:05:25 JST 2016	a["{\"msg\":\"added\",\"collection\":\"users\",\"id\":\"wffqrSXNojgMnuz8v\",\"fields\":{\"emails\":[{\"address\":\"yusuke.iwaki@example.com\",\"verified\":true}],\"username\":\"YusukeIwaki\"}}"]

Thu Dec 08 20:05:25 JST 2016	Thu Dec 08 20:05:25 JST 2016	a["{\"msg\":\"result\",\"id\":\"1\",\"result\":{\"id\":\"wffqrSXNojgMnuz8v\",\"token\":\"0JdBqCN90XpH・・・・・・・dB_Ti7L3bB0\",\"tokenExpires\":{\"$date\":1488897869020}}}"]
```

When I observed DDP messages last month, each session observer was triggered after user login.


By the way, I couldn't understand why `fetchPublicSettings` and `registerForPush` are required to execute before user login... Could you explain?

app/src/main/java/chat/rocket/android/activity/ServerConfigActivity.java
laggedHero: And without the change to the query in the observer (around line 57) this block is called a lot of times.... like in a loop forever. So I changed it in order to stop that behavior. But it will be called a couple of times.

app/src/main/java/chat/rocket/android/activity/ServerConfigActivity.java
laggedHero: Public settings contain some settings that may (will) be used by the app and may be applied before the login.

And by registering for push before the login we keep up with the current behavior already in place in the Cordova app and which could be specially useful on the white labels.

YusukeIwaki: I see.

I think it is better to put `fetchPublicSettings` before registerListeners in RocketChatWebSocketThread, and create new observer (PushSettingsObserver) to observe whether push  is enabled in public settings and trigger `registerForPush` if the enabled setting is observed.

app/src/main/java/chat/rocket/android/activity/ServerConfigActivity.java
YusukeIwaki: Hmm, I think it is better not to mix other bug-fixes inrelevant to push-notification feature...

laggedHero: This is far from `inrelevant`. This will generate a loop of requests and will trigger a lot of push requests.

YusukeIwaki: The loop is mainly caused by calling `registerForPush` in sessionObserver's callback.
It would be fixed by moving `registerForPush` out of the Activity as mentioned in https://github.com/RocketChat/Rocket.Chat.Android/pull/124#discussion_r94358779

YusukeIwaki: The loop is also caused by a bug of RealmObjectObserver, which is fixed in https://github.com/RocketChat/Rocket.Chat.Android/pull/127
(my fault,,, sorry...)

This would also fix the loop problem.

laggedHero:
@YusukeIwaki looks better?

YusukeIwaki:
I will refactor some classes after merging this.

YusukeIwaki (approved):

PR #126, opened 2017-01-03
Added a burger menu

### LINKED ISSUE #97: [v1.0-beta] Add burger menu
We should add the burger menu to the app and have the room type icon next to the room name.

Example

screen shot 2016-12-21 at 11 55 50 am

laggedHero:
Uses black color (easy to tint later)
Fixes tinting in the message action selector

@RocketChat/android

Closes #97 

People are used to the burger menu. So I'm adding one in the app.


YusukeIwaki:
Why not modify HomeFragment#onResume

`activityToolbar.setNavigationIcon(null);`

to

`activityToolbar.setNavigationIcon(R.drawable.ic_menu_black_24dp);` ?

This pull request seems to work well, but too complicated just for showing burger menu.

laggedHero:
Because `We should add the burger menu to the app and have the room type icon next to the room name.` as stated in #97

YusukeIwaki:
Hmm,  Toolbar is not designed to have 2 icons.
So it is better to define custom view extending Toolbar with an extra icon in `rocket-chat-android-widgets` project.

laggedHero:
Hmm. That's a nice idea. I'll probably push the changes tomorrow.

laggedHero:
@YusukeIwaki could you please take a look? Thx!

geekgonecrazy:
I don't know if this is intentional.  But the server menu that you swipe left when the channel list is expanded no longer comes out.  I just discovered it was there in the last build from develop, and noticed in this one its gone.

Figured i'd point it out just in case a regression :)

laggedHero:
@geekgonecrazy I'll take a look here. Anyway, we need to find a way to unearth that menu and make it clear it exists.

laggedHero:
Not a regression per se: it will depend on how close to the border you try to swipe. We are using a slide inside a slide... so it may be something.

Let's leave it like that and figure out a secondary way to expose that menu?

geekgonecrazy:
:+1: works just fine.Just slid from mid way instead of at very edge.  But I agree needs to be more discoverable.

rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/CustomToolbar.java
```diff
@@ -0,0 +1,98 @@
...
+import android.support.annotation.Nullable;
+import android.support.annotation.StringRes;
+import android.support.graphics.drawable.VectorDrawableCompat;
+import android.support.v4.graphics.drawable.DrawableCompat;
+import android.support.v7.widget.Toolbar;
+import android.util.AttributeSet;
+import android.view.View;
+import android.widget.TextView;
+
+public class CustomToolbar extends Toolbar {
```
YusukeIwaki: `Custom` is not good name.
Please use RoomToolbar or another one to indicate it should be used in "Room".

app/src/main/java/chat/rocket/android/activity/MainActivity.java
```diff
@@ -91,25 +94,48 @@ private void setUserAwayIfServerAvailable() {
...
+    toolbar.setNavigationOnClickListener(view -> {
+      if (pane.isSlideable() && !pane.isOpen()) {
+        pane.openPane();
+      }
+    });
+  }
+
+  private void setupToolbar() {
+    Toolbar toolbar = (Toolbar) findViewById(R.id.activity_main_toolbar);
+    setSupportActionBar(toolbar);
```
YusukeIwaki: I think it is better not to call `setSupportActionBar`. 
Because now we use very customized Toolbar, it is not good for the toolbar to be customized unexpectedly by Android framework.

Just put a customized Toolbar and setup it in each RoomFragment.

rocket-chat-android-widgets/src/main/res/values/custom_toolbar_attrs.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <declare-styleable name="CustomToolbar">
+        <attr name="titleText" format="string" />
+        <attr name="titleTextColor" format="color" />
```
YusukeIwaki: it is better to set `titleTextAppearance` rather than only color.

rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/CustomToolbar.java
```diff
@@ -0,0 +1,89 @@
...
+  @Override
+  public void setTitle(CharSequence title) {
+    if (titleTextView != null) {
+      titleTextView.setText(title);
+      return;
+    }
+    super.setTitle(title);
+  }
+
+  public void setTitleDrawableLeft(@DrawableRes int drawableResId) {
```
YusukeIwaki: `DrawableLeft` is not good.
`setRoomTypeIcon` is better.

YusukeIwaki:
Dual sliding pane is just a temporal solution... 😓 
It should be replaced more discoverable one when we implement "connect to multiple server" functionality.

YusukeIwaki:
@laggedHero I want to try #134. It is really related to this PR. Can I take also this PR?

laggedHero:
It would be better if I finish this one and then you make the other. They're two diferent matters and should be treated as such.

I'll clean up this and we'll be able to merge it.

rocket-chat-android-widgets/src/main/res/values/custom_toolbar_attrs.xml
laggedHero: We are not even using text appearance. Let's add it when we need it.

YusukeIwaki (approved):
👍

YusukeIwaki:
Sure, let's merge this PR first.

PR #127, opened 2017-01-04
fix the issue that RealmObjectObserver's callback is called repeatedly

YusukeIwaki:
@RocketChat/android

When the result of the query is null, RealmObjectObserver calls the callback function repeatedly with `null` parameter because

```
      if (previousResultsString != null && previousResultsString.equals(currentResultString)) {
        return;
      }
```

When previousResultsString = null, currentResultString = null,
not returned here...

`previousResultsString != null` is necessary for initial callback. 
So I replaced `currentResultString = null` to ` currentResultString = ""`


laggedHero:
Why not initialize the `previousResultsString` with an empty string? No need for a null check then. xD

YusukeIwaki:
If previousResultsString is initialized with `""`, initial callback is never called when the result is null.
It is not good.

So I left `previousResultsString = null` to check if it is the first call.

laggedHero:
I see. Looks good.

laggedHero (approved):

PR #130, opened 2017-01-09
Refactor push-notification related codes

YusukeIwaki:
@RocketChat/android

Naming and logical improvement for #124.

* rename some classes
* rename some methods
* [x] remove MethodCallHelper(String serverConfigId) usage
* [x] move pushSetUser to SessionObserver#onLogin
* ~remove PushInteractor~
* [x] move `syncPushToken` from ServerConfig to PushConfig

slice


app/src/main/java/chat/rocket/android/api/MethodCallHelper.java
```diff
@@ -39,13 +39,6 @@
   protected final DDPClientWrapper ddpClient;
 
   /**
-   * initialize with ServerConfigId.
-   */
-  public MethodCallHelper(String serverConfigId) {
```
YusukeIwaki: This method is really not intended to use. (I used to mark as `@deprecated`)
https://github.com/RocketChat/Rocket.Chat.Android/pull/124/files#diff-6e3f31a463d9db5868ae9b899f882c42

app/src/main/java/chat/rocket/android/push/interactors/DefaultPushInteractor.java
```diff
@@ -1,19 +0,0 @@
-package chat.rocket.android.push.interactors;
-
-import chat.rocket.android.helper.ServerPolicyHelper;
-import chat.rocket.android.model.ServerConfig;
-import chat.rocket.android.realm_helper.RealmStore;
-
-public class DefaultPushInteractor implements PushInteractor {
```
YusukeIwaki: The role of `PushInteractor` is not clear for now.
So I removed it just now.

laggedHero: The idea is to make `PushNotificationHandler` unaware of `Realm` stores and etc. On the long run, we want `Realm` to be an implementation detail and not something that's everywhere in the app.

So `PushNotificationHandler` only knows about a `PushInteractor` interface and the `DefaultPushInteractor` implements that using `Realm`.

laggedHero (changes requested):
Don't remove the interactor. Improve it. :)

app/src/main/java/chat/rocket/android/push/interactors/DefaultPushInteractor.java
YusukeIwaki: I understood what you intended.

However I think the PushInteractor interface is not well-encapsulated.
I will try another improvement.

laggedHero: And it's not.

It would be better if we have a `Repository` interface and an implementation of that repository using `Realm`.

Another improvement would be making our domain entities not extend `RealmObject` at all.

We could list a lot of improvements but we need to do things a step at a time.

If you want to take the next step, please do it. :)

YusukeIwaki: Sorry for posting incomplete comment...

OK I will write a list first.

app/src/main/java/chat/rocket/android/service/observer/CurrentUserObserver.java
```diff
@@ -22,15 +20,13 @@
  */
 public class CurrentUserObserver extends AbstractModelObserver<User> {
```
YusukeIwaki, line 21: CurrentUserObserver is just a subscription of "activeUser" (just observing a user with email) and actually unaware of the state if the "login" operation is finished or not, while SessionObserver knows when login completed.

laggedHero (approved):

PR #146, opened 2017-01-12
Using diff util to update the list

laggedHero:
@RocketChat/android

Won't reload the entire message list when a new message show up

### DIFF
--- a/app/src/main/java/chat/rocket/android/layouthelper/ExtRealmModelListAdapter.java
+++ b/app/src/main/java/chat/rocket/android/layouthelper/ExtRealmModelListAdapter.java
@@ -3,7 +3,6 @@ package chat.rocket.android.layouthelper;
 import android.content.Context;
 import android.support.annotation.LayoutRes;
 import android.support.v7.util.ListUpdateCallback;
-import android.support.v7.widget.RecyclerView;
 import io.realm.RealmObject;
 
 import chat.rocket.android.realm_helper.RealmModelListAdapter;
--- a/app/src/main/java/chat/rocket/android/layouthelper/ExtRealmModelListAdapter.java
+++ b/app/src/main/java/chat/rocket/android/layouthelper/ExtRealmModelListAdapter.java
@@ -17,8 +17,6 @@ public abstract class ExtRealmModelListAdapter<T extends RealmObject, VM,
   protected static final int VIEW_TYPE_HEADER = -1;
   protected static final int VIEW_TYPE_FOOTER = -2;
 
-  private UpdateListener updateListener;
-
   protected ExtRealmModelListAdapter(Context context) {
     super(context);
   }
@@ -40,17 +38,10 @@ public abstract class ExtRealmModelListAdapter<T extends RealmObject, VM,
     notifyItemChanged(position + 1);
   }
 
-  public void setUpdateListener(UpdateListener updateListener) {
-    this.updateListener = updateListener;
-  }
-
   protected ListUpdateCallback listUpdateCallback = new ListUpdateCallback() {
     @Override
     public void onInserted(int position, int count) {
       notifyItemRangeInserted(position + 1, count);
-      if (updateListener != null) {
-        updateListener.onInserted(count);
-      }
     }
 
     @Override
@@ -119,11 +110,5 @@ public abstract class ExtRealmModelListAdapter<T extends RealmObject, VM,
   @Override
   public ListUpdateCallback getListUpdateCallback() {
     return listUpdateCallback;
-  }
-
-  // We'll be using the insert event only as of now
-  // Let's add more events when/if needed
-  public interface UpdateListener {
-    void onInserted(int count);
   }
 }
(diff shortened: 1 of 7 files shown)


app/src/main/java/chat/rocket/android/layouthelper/ExtRealmModelListAdapter.java
```diff
@@ -37,6 +41,35 @@ protected void notifyRealmModelItemChanged(int position) {
...
+    this.layoutManager = layoutManager;
+  }
+
+  protected ListUpdateCallback listUpdateCallback = new ListUpdateCallback() {
+    @Override
+    public void onInserted(int position, int count) {
+      notifyItemRangeInserted(position + 1, count);
+      if (layoutManager != null) {
+        layoutManager.scrollToPosition(0);
+      }
```
YusukeIwaki: I will check this behavior...

YusukeIwaki:
Awesome! 👍 👍 

I will check your code and APK.

app/src/main/java/chat/rocket/android/layouthelper/ExtRealmModelListAdapter.java
YusukeIwaki: I suspect this causes unexpected scroll.
When we read very past message, it is better not to force scroll to the bottom by new message arrival.

Also, I think it is better for this kind of functionality to be implemented with RealmListObserver like this in RoomFragment.

```
latestMessageObserver =  realmHelper
        .createObjectObserver(
            realm -> realm.where(Message.class).equalTo("rid", roomId))
        .setImpl(results -> results.last(null))
        .setOnUpdateListener(this::scrollToBottomIfNeeded);
```

rather than passing layouthelper

YusukeIwaki:
Awesome!! 👍 👍 

However just https://github.com/RocketChat/Rocket.Chat.Android/pull/146#discussion_r96007439 want to be fixed.

app/src/main/java/chat/rocket/android/layouthelper/ExtRealmModelListAdapter.java
laggedHero: I'll do a fix for the behavior. But I'm not adding another hard dependency on `Realm`. We really need to stop doing it.

YusukeIwaki: My concern is that it is better to implement with "observer" here.

slice

For example this kind of improvement is hard to implement with extending RecyclerView adapter.

laggedHero: I'm adding a listener that will be called back.

```
  // We'll be using the insert event only as of now
  // Let's add more events when/if needed
  public interface UpdateListener {
    void onInserted(int count);
  }
```

`RoomFragment` may do wherever it feels like. And since it calls back with the total new inserted... should enable your use case.

laggedHero: If listeners are not your thing (some ppl don't like it) I could change it for some Rx based stuff.

But not another hard `Realm` dependency.

YusukeIwaki: Shall we discuss the feature of autoscroll (or notify new message) in another Pull Request?

This Pull Request is already really worthy just for adding DiffUtils base codes.

laggedHero: Yep. We could do that. But not adding a new hard `Realm` dependency is the way to go. :)

laggedHero:
I don't have the permissions on Travis to run the test again.

I'll run the CircleCI one.

app/src/main/java/chat/rocket/android/layouthelper/ExtRealmModelListAdapter.java
YusukeIwaki: I agree with avoiding the hard Realm dependency.
I actually prefer the method with Rx like your suggestion 😅 .

Anyway, shall we separate the stuff other than related to DiffUtils here?
https://github.com/RocketChat/Rocket.Chat.Android/pull/146/commits/ae5211eb3b4c64852bc555e3b28ff1d2803ef849
I created a commit to remove it. Could you provide a new PR with reverting the commit?

YusukeIwaki (approved):

laggedHero:
@YusukeIwaki simply reverting things is pretty rude. A PR is a place for discussing.

I'll shift my focus to the other projects for a while.

YusukeIwaki:
@laggedHero Sorry for unkindness, as I commented here https://github.com/RocketChat/Rocket.Chat.Android/pull/146#discussion_r96034061 in advance, I expected you to create new Pull Request for discussing the feature of autoscroll (or notify new message) with reverting the commit https://github.com/RocketChat/Rocket.Chat.Android/pull/146/commits/ae5211eb3b4c64852bc555e3b28ff1d2803ef849.

PR #149, opened 2017-01-14
show budge for unread message count

### LINKED ISSUE #134: Indicator of channels with unread messages
Related to #126 and #97 it would be useful to have an indicator that there are unread messages in other channels.  As well as a count on how many mentions.   

This is what we have on our web view and our cordova app:

image

YusukeIwaki:
@RocketChat/android

Closes #134 

add indicator into the burger icon for notifying the count of unread room subscriptions.

Rocket.Chat (webapp) originally shows "999+" badge when the number of mentions > 999. However "999+" cannot be layout on a small badge. So "99+" is used instead in this app.

### DIFF
--- a/rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/RoomToolbar.java
+++ b/rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/RoomToolbar.java
@@ -2,13 +2,13 @@ package chat.rocket.android.widget;
 
 import android.content.Context;
 import android.content.res.TypedArray;
-import android.graphics.Color;
 import android.graphics.Typeface;
 import android.graphics.drawable.Drawable;
 import android.support.annotation.DrawableRes;
 import android.support.annotation.Nullable;
 import android.support.annotation.StringRes;
 import android.support.graphics.drawable.VectorDrawableCompat;
+import android.support.v4.content.ContextCompat;
 import android.support.v4.widget.TextViewCompat;
 import android.support.v7.widget.AppCompatImageView;
 import android.support.v7.widget.Toolbar;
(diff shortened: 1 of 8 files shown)


geekgonecrazy:
On just testing this one, it looks like its incrementing the number based on number of unread channels?  We might want to adjust this though to have the count only present for number of mentions / direct messages.  

Because mentioned are generally the ones a user is going to care about.  Like for instance on the demo server.  I have 8 channels with un-read messages in it.  Most of those channels I view and attack with purpose when I have the extra time to properly pay attention and help users out.  But if I know I have a direct mention or message.  I will jump right to it and knowing that at a glance is very useful.

YusukeIwaki:
@geekgonecrazy Thank you for your feedback!
I've just understand which number would be shown on the budge.

I will adjust the behavior.

laggedHero:
@geekgonecrazy Does this update works as expected?

geekgonecrazy:
@laggedHero checking now!

geekgonecrazy:
@laggedHero appears to! :+1:

rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/RoomToolbar.java
```diff
@@ -95,4 +104,68 @@ public void setRoomIcon(@DrawableRes int drawableResId) {
...
+      try {
+        Field field = Toolbar.class.getDeclaredField("mNavButtonView");
+        field.setAccessible(true);
+        ImageButton navButtonView = (ImageButton) field.get(this);
+        int iconLeft = navButtonView.getLeft();
+        int iconTop = navButtonView.getTop();
+        int iconRight = navButtonView.getRight();
+        int iconBottom = navButtonView.getBottom();
+
+        int budgeLeft = iconLeft + (iconRight - iconLeft) * 5 / 8;
```
laggedHero: Could you please add some explanation for these maths? Thx

YusukeIwaki: Sure. And I will also fix typo `budge` 😓

laggedHero (approved):

PR #151, opened 2017-01-15
auto scroll to latest message

YusukeIwaki:
originally proposed in #146 (Thanks to laggedHero! 🙇 )

---

After DiffUtils is intriduced in #146, message list view is not automatically scrolled when new message is arrived. (ref: https://code.google.com/p/android/issues/detail?id=174227 )

* Automatically scroll to the new message  if the scroll position is near the new message
* show indicator if user is reading very past message not to disturb reading

### DIFF
--- a/app/src/main/java/chat/rocket/android/fragment/chatroom/RoomFragment.java
+++ b/app/src/main/java/chat/rocket/android/fragment/chatroom/RoomFragment.java
@@ -6,6 +6,7 @@ import android.content.Intent;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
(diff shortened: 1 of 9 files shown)


laggedHero:
Something is forcing the views to re-render. I'll try to find out what it is.

laggedHero:
@YusukeIwaki the `User` on the `Message` have some objects that trigger a diff.

I tried implementing (with the help of AS) `equals` and `hashCode`. That worked out (and should be a better solution overall).

YusukeIwaki:
@laggedHero Thank you for your investigation 👍 

Could you provide the change by creating another pull request?

laggedHero:
Yep. That works.

laggedHero (approved):

Issue #147, opened 2017-01-13
[v1.0-beta] alternative WebSocket connectivity management


YusukeIwaki:
Managing websocket connectivity with ServerConfig  is not good.
It is really complicated to handle every usecase with mapping it to a exclusive state, and really buggy now... (#117, #137, #142, ...)

Let's introduce clean architecture partially just for managing the WebSocket connectivities.

---
![slice](https://cloud.githubusercontent.com/assets/11763113/22061001/63c044d4-ddb6-11e6-837e-f080e0eaf88f.png)

geekgonecrazy:
Might also be worth investigating the rest api in some instances.  Like if a websocket is not established and a message is sent.  Waiting on a websocket might not best.  Might be good to just fire of quick http request and be done with it.  Just a thought

YusukeIwaki:
> Like if a websocket is not established and a message is sent. Waiting on a websocket might not best.

Yes, exactly.
A piece of REST API code seems introduced in #153 and I expect them to be extended in future.

laggedHero:
I've just closed the other issues since everything seems to be related. :)

geekgonecrazy:
:+1: makes sense.  For sure what they are all related to.

PR #153, opened 2017-01-16
App connects to insecure servers.

### LINKED ISSUE #148: Unable to test on local server
The check server code is adding a `https` in front of the server domain (you can see [here](https://github.com/RocketChat/Rocket.Chat.Android/blob/develop/app/src/main/java/chat/rocket/android/helper/ServerPolicyHelper.java#L33)) and because of this I'm unable to connect on my local meteor app.

laggedHero:
@RocketChat/android

Closes #148

The app may now connect to insecure servers.

Should we warn the user?

### DIFF
--- a/app/src/main/java/chat/rocket/android/api/rest/ServerPolicyApi.java
+++ b/app/src/main/java/chat/rocket/android/api/rest/ServerPolicyApi.java
@@ -1,19 +1,15 @@
 package chat.rocket.android.api.rest;
 
-import android.support.annotation.NonNull;
 import org.json.JSONObject;
+
+import rx.Observable;
 
 public interface ServerPolicyApi {
 
-  void getApiInfoSecurely(@NonNull Callback callback);
+  String SECURE_PROTOCOL = "https://";
+  String INSECURE_PROTOCOL = "http://";
 
-  void getApiInfoInsecurely(@NonNull Callback callback);
+  Observable<Response<JSONObject>> getApiInfoSecurely();
 
-  interface Callback {
-    void onSuccess(JSONObject jsonObject);
-
-    void onResponseError();
-
-    void onNetworkError();
-  }
+  Observable<Response<JSONObject>> getApiInfoInsecurely();
 }
--- a/app/src/main/java/chat/rocket/android/api/rest/ServerPolicyApi.java
+++ b/app/src/main/java/chat/rocket/android/api/rest/ServerPolicyApi.java
@@ -1,0 +1,19 @@
+package chat.rocket.android.api.rest;
+
+import android.support.annotation.NonNull;
+import org.json.JSONObject;
+
+public interface ServerPolicyApi {
+
+  void getApiInfoSecurely(@NonNull Callback callback);
+
+  void getApiInfoInsecurely(@NonNull Callback callback);
+
+  interface Callback {
+    void onSuccess(JSONObject jsonObject);
+
+    void onResponseError();
+
+    void onNetworkError();
+  }
+}
--- a/app/src/main/java/chat/rocket/android/helper/ServerPolicyHelper.java
+++ b/app/src/main/java/chat/rocket/android/helper/ServerPolicyHelper.java
@@ -4,7 +4,7 @@ import android.support.annotation.NonNull;
 
 import org.json.JSONObject;
 
-import chat.rocket.android.api.rest.ServerPolicyApi;
+import rx.Observable;
 
 public class ServerPolicyHelper {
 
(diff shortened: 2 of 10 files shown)


YusukeIwaki:
I think it would be better to limit connecting without secure connection, 
for example, only available on debug build or white-label build.

Functionality seems OK 👍  with your codes.

app/src/main/java/chat/rocket/android/helper/ServerPolicyHelper.java
```diff
@@ -104,9 +92,64 @@ private static boolean isVersionValid(String version) {
...
+  }
+
   public interface Callback {
-    void isValid();
+    void isValid(boolean usesSecureConnection);
 
     void isNotValid();
+
+    void onNetworkError();
   }
```
YusukeIwaki, line 109: It would be better to define a class for validation like `ServerPolicyVaridator` because secure/insecure validation has very similar logic.

laggedHero:
Since people may deploy their own instance of RC... it means that we may have insecure connections. :/

While I see the case for blocking it out, we cannot enforce it.

YusukeIwaki:
I see...

It would be good to make confirmation before failover to http connection, I think.

YusukeIwaki:
@laggedHero sorry, let me take back last comment...

To make confirmation before connecting https, it would make the logic very complex.
So I think it is better to merge this pull request first without it, and some improvement will be wanted after #147 resolved.

In #147, I plan to remove ServerConfig partially and replace some codes with usecase-based classes. So improvement can be made more easily after that.

geekgonecrazy:
I think it might be worth in the future doing something of that nature.  But I think the notice mainly needs shown to the admin, with a link to our documentation on adding ssl. Otherwise the layman user will have no idea.

laggedHero:
@geekgonecrazy Warning the admin only may be a nice thing. Don't we have something like that in place for the web app?

Maybe a warning on the home or something. That would be nice as a platform-wide feat.

Don't you think so?

geekgonecrazy:
@laggedHero yeah I agree it should probably be something that is triggered server side to just alert the admin in general.  Maybe like we do if the site_url is different then initial.

YusukeIwaki (approved):
👍

app/src/main/java/chat/rocket/android/api/rest/ServerPolicyApi.java
```diff
@@ -0,0 +1,19 @@
...
+import android.support.annotation.NonNull;
+import org.json.JSONObject;
+
+public interface ServerPolicyApi {
+
+  void getApiInfoSecurely(@NonNull Callback callback);
+
+  void getApiInfoInsecurely(@NonNull Callback callback);
+
+  interface Callback {
```
YusukeIwaki: This callback interface would be shared for all REST APIs. So it is better to define isolatedly.

YusukeIwaki: Also, I think `rx.Single` or  `bolts.Task` would make it easier to write the logic of sequential API calls or retrying, and actually prefer using them.
How do you think?

laggedHero: When/if we get a second REST call then that would be the right moment to extract it to its own class. This will probably be our only call outside the file upload stuff.

And we're discussing if we should create a method to obtain the same set of data.

About the `Rx` / `Bolts` stuff: that may work. I'll build something here.

YusukeIwaki:
@laggedHero Thank you for trying rx.Observable for REST API.

I checked the code, however I'm afraid that the previous version (with Callback class) is easier to read...
Maybe it is better for us  to create RestAPIClient class in introducing Rx and capsulize OkHttp client method calls. Shall we discuss it with another issue or pull request and keep this pull request focusing on "connecting via http/ws" feature?

 I actually also want to try it (after #147 is fixed... 😓  ) and introduce what I've tried in [another project](https://github.com/YusukeIwaki/GitHub-Issue-Viewer-2/blob/develop/app/src/main/java/io/github/yusukeiwaki/githubviewer2/webapi/ApiBase.java) there.

laggedHero:
I forgot to unsubscribe... so there was a memory leak. Fixed.

About the easy to read: I think the new version is a lot easier.

About encapsulating `OkHttp` calls: maybe we should think about using `Retrofit`. And even if we're not using it we may encapsulate it with `Rx`. We may probably be able to remove `bolts` from the project by using `Rx` alone.

If you don't mind I'll merge this PR. We may make more changes by the time we figure out where we're heading to.

PR #154, opened 2017-01-16
Implemented equals and hashCode

laggedHero:
Using IDE helpers

@RocketChat/android

I chose to implement it in the safest way (checking for null everywhere). Maybe that could be changed.


YusukeIwaki:
Great work! 👍 

By the way, do you have any good library or method to generate these code with the annotation processor like [EqualsAndHashCode](https://projectlombok.org/features/EqualsAndHashCode.html) or [auto-value](https://github.com/remind101/auto-value-realm) or any other similar one?

It would better for us to replace raw getter/setter/equals/hashCode... with some kind of auto-generating method, in future.

YusukeIwaki (approved):

laggedHero:
I think that as soon as we get `Realm` out of the `core`, we may use `auto-value`.

:)

PR #160, opened 2017-01-18
Using a dialog fragment

### LINKED ISSUE #143: Attachment options UI not full width
Your Rocket.Chat.Android version: alpha from Jan 11

This is how it looks on Android 6.0.1:

laggedHero:
@RocketChat/android

Closes #143


rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/internal/ExtraActionPickerDialogFragment.java
```diff
@@ -0,0 +1,62 @@
...
+
+    return fragment;
+  }
+
+  public void setActionItems(List<MessageExtraActionItemPresenter> actionItems) {
+    this.actionItems = actionItems;
+  }
+
+  public void setCallback(Callback callback) {
+    this.callback = callback;
```
YusukeIwaki: Does this work when Fragment is re-instantiated by Android framework...?

I'm afraid that I don't know the detail, but it seems that Google doesn't recommend creating setter methods in Fragment for setting custom callbacks.
https://developer.android.com/training/basics/fragments/communicating.html

rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/message/MessageFormLayout.java
```diff
@@ -159,23 +151,11 @@ public void run() {
...
-
-  public interface ActionListener {
-    void onSubmitText(String message);
+  public interface ShowExtraActionSelectionCallback {
+    void show();
+  }
 
-    void onExtraActionSelected(int itemId);
+  public interface TextListener {
+    void onSubmitText(String message);
```
YusukeIwaki, line 158: I think it is better to include the word `Submit` in the name of interface, like `SubmitTextCallback`

rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/message/MessageFormLayout.java
```diff
@@ -110,17 +101,18 @@ public void afterTextChanged(Editable s) {
...
+    this.showExtraActionSelectionCallback = showExtraActionSelectionCallback;
+  }
 
-    if (!found) {
-      extraActionItems.add(itemPresenter);
+  public void setTextListener(TextListener textListener) {
+    this.textListener = textListener;
+  }
+
+  private void showExtraActionSelectionDialog() {
```
YusukeIwaki: If MessageFormLayout doesn't show Dialog directly, this class should not know that the callback is used for showing dialog. So it is not better to include the word `dialog` in the method and callback interface.

I think it is better to name this method and the callback interface simply like `OnExtraActionSelectionButtonClicked` and `setOnExtraActionSelectionButtonClickedListener` . (too long 😅 ... more shorter one is prefered :) )

app/src/main/java/chat/rocket/android/fragment/chatroom/RoomFragment.java
```diff
@@ -295,9 +301,15 @@ private boolean closeSideMenuIfNeeded() {
   }
 
   private void setupMessageComposer() {
+    extraActionItems = new ArrayList<>(3); // fixed number as of now
+    extraActionItems.add(new ImageUploadActionItem());
+    extraActionItems.add(new AudioUploadActionItem());
+    extraActionItems.add(new VideoUploadActionItem());
```
YusukeIwaki: This is not used in MessageFormLayout. So it would be better to set in `showExtraActionSelectionDialog`.

rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/internal/ExtraActionPickerDialogFragment.java
laggedHero: Oh.. I forgot that. And I can see why that could be troublesome.

Thx

app/src/main/java/chat/rocket/android/renderer/UserRenderer.java
```diff
@@ -19,7 +19,7 @@ public UserRenderer(Context context, User user) {
   }
 
   /**
-   * show Avatar image.
+   * onClick Avatar image.
```
YusukeIwaki: It seems that some comments are unexpectedly renamed... 😅

YusukeIwaki (approved):
Great 👍

app/src/main/java/chat/rocket/android/renderer/UserRenderer.java
laggedHero: :O

I'll fix it

YusukeIwaki: Thank you! :)

Issue #137, opened 2017-01-11
Issues resuming from background


geekgonecrazy:
Still working to identify this.  But when resuming from background, often it does not seem to get latest on the current channel in focus.

For example, if i'm having a one on one conversation and put in the background and I get a notification of a new message and restore from background (not tap notification.  That action does seem to work) the latest messages are not loaded.

This seems to happen more when its been backgrounded longer.  Usually to get around this I swipe it away from recents and re-open.  Most of the time this does the trick.  Others I've had to force stop and then open again.

Its un-clear if it simply hasn't established a connection to the server again, or is just failing to load those missed messages.

Might be related to #117 its hard for me to be certain

YusukeIwaki:
It is caused by the state mismatch of ServerConfig model and RocketChatWebSocketThread...
By killing the RocketChatService manually (swipe away from recent tasks), the state is reset and works as expected.

It is really should be fixed..

cheonjoosung:
I tested this issue (related #117) both demo.rocket.chat server & my own rocket server. Surpringly, the problem didn't happen in demo.rocket.chat server almostly.
But, the problem happen frequently in my own server.

YusukeIwaki:
The root cause is that `RxWebSocket#isConnected()` returns true unexpectedly, even after the WebSocket connection is disconnected silently for some reason...

https://github.com/RocketChat/Rocket.Chat.Android/blob/develop/android-ddp/src/main/java/chat/rocket/android_ddp/rx/RxWebSocket.java

laggedHero:
We have an issue to redo the connectivity. Please follow #147

geekgonecrazy:
I know #170 is where redirected to.  But it was closed and still seeing this

Tons of issues resuming from background and it not updating channel until I switch to a new channel and back, also going on mobile network where signal is not great.

I'll be in DM with a user, send to background get a push notification, resume and the message isn't there, have to switch channels and then back and message appears.  

Other times will be in conversation and they will send a message and it'll show in sidebar a new message but won't show in channel pane.  I have to again switch channels and back to get it to show up.

laggedHero:
@YusukeIwaki do you think you may that care of this one?

YusukeIwaki:
```
3:54:49.169 V/RocketChatService( 2286): ⇢ getOrCreateWebSocketThread(hostname="demo.rocket.chat") [Thread:"RxComputationScheduler-4"]
[...]

PR #170, opened 2017-01-20
introduce alternative WebSocket connectivity management

### LINKED ISSUE #147: [v1.0-beta] alternative WebSocket connectivity management

YusukeIwaki:
@RocketChat/android

Closes #147 

### ServerConfig is removed! 

* Now `ConnectivityManager` is responsible for the connectivity management.
* `serverConfigId` is no longer used, `hostname` is used as a primary key for each model.

### DDPClient#isConnected is removed!

`DDPClient#isConnected` returns true often even after connection is disconnected. It was really confusing! ( #137 is caused by it)
Now the function for checking connectivity is implemented with DDP ping-pong. It costs a little time, however it brings much more precise connectivity check.

---

### about RxJava usage...

I couldn't find `flatMap` or its compatible operator in `rx.Completable` .
( `andThen` looks similar but it doesn't evaluate item with laziness... ) So I used `Single<Boolean>` instead.


app/src/main/java/chat/rocket/android/helper/RxHelper.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+public class RxHelper {
+  public static Func1<Observable<? extends Throwable>, Observable<?>> exponentialBackoff(
+      int maxRetryCount, long base, TimeUnit unit) {
+    return errors -> errors
+        .zipWith(Observable.range(0, maxRetryCount), (error, retryCount) -> retryCount)
+        .flatMap(retryCount -> Observable.timer(base * (long) Math.pow(2, retryCount), unit));
+  }
+
+  public static <T> Single<T> lazy(Func0<Single<T>> func) {
```
laggedHero: Won't `rx.Defer` be enough?

app/src/main/java/chat/rocket/android/helper/RxHelper.java
```diff
@@ -0,0 +1,23 @@
...
+import rx.functions.Func1;
+
+/**
+ * Rx operator and so on.
+ */
+public class RxHelper {
+  public static Func1<Observable<? extends Throwable>, Observable<?>> exponentialBackoff(
+      int maxRetryCount, long base, TimeUnit unit) {
+    return errors -> errors
+        .zipWith(Observable.range(0, maxRetryCount), (error, retryCount) -> retryCount)
```
laggedHero, line 16: Why `zipWith` if you're ignoring the `error`?

app/src/main/java/chat/rocket/android/service/RocketChatWebSocketThread.java
```diff
@@ -123,111 +137,149 @@ private void forceInvalidateTokens() {
...
+      super.quit();
+      return Single.just(true);
     }
   }
 
   /**
+   * THIS METHOD THROWS EXCEPTION!!
+   * Use terminate() instead!!
+   */
+  @Deprecated public final boolean quit() {
```
laggedHero: `@Override`?

app/src/main/java/chat/rocket/android/service/ServerInfoImpl.java
```diff
@@ -0,0 +1,98 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import chat.rocket.android.realm_helper.RealmHelper;
+import chat.rocket.android.realm_helper.RealmStore;
+
+/**
+ * Backend implementation to store ServerInfo.
+ */
+public class ServerInfoImpl extends RealmObject {
```
laggedHero: This is not an implementation of `ServerInfo`... but a `Realm` version of it. Maybe `RealmServerInfo`?

app/src/main/java/chat/rocket/android/service/ConnectivityManagerImpl.java
```diff
@@ -0,0 +1,230 @@
...
+import chat.rocket.android.log.RCLog;
+import hugo.weaving.DebugLog;
+import rx.Observable;
+import rx.Single;
+import rx.subjects.PublishSubject;
+
+/**
+ * Connectivity management implementation.
+ */
+/*package*/ class ConnectivityManagerImpl implements ConnectivityManagerApi, ConnectivityManagerInternal {
```
laggedHero: This `ConnectivityManagerImpl` talks directly with `ServerInfoImpl` (which is really a `RealmServerInfo`).

So maybe `RealmConnectivityManager` would be a better name.

app/src/main/java/chat/rocket/android/LaunchUtil.java
```diff
@@ -23,10 +33,10 @@ public static void showAddServerActivity(Context context) {
   /**
    * launch ServerConfigActivity with proper flags.
    */
-  public static void showLoginActivity(Context context, String serverCondigId) {
+  public static void showLoginActivity(Context context, String hostname) {
     Intent intent = new Intent(context, LoginActivity.class);
     intent.setFlags(Intent.FLAG_ACTIVITY_REORDER_TO_FRONT | Intent.FLAG_ACTIVITY_CLEAR_TOP);
-    intent.putExtra("serverConfigId", serverCondigId);
+    intent.putExtra("hostname", hostname);
```
laggedHero: What about a constant on `LaunchUtil`?

laggedHero (commented):
As a rule of thumb `Impl` should be avoided: it usually means we have some bad design at hand.

app/src/main/java/chat/rocket/android/helper/RxHelper.java
YusukeIwaki: Actually I made this using [javadoc](https://github.com/ReactiveX/RxJava/blob/a8ba158839b67246a742b6f1531995ffd7545c08/src/main/java/io/reactivex/Observable.java#L9601) as reference, and didn't know the detail...
I will study it...

app/src/main/java/chat/rocket/android/service/RocketChatWebSocketThread.java
YusukeIwaki: This method is really not intended to use.
So I want to add some kind of annotation for indicating it.

app/src/main/java/chat/rocket/android/service/ServerInfoImpl.java
YusukeIwaki: Yes, exactry!
That sounds nice. I will fix.
Thank you.

app/src/main/java/chat/rocket/android/service/RocketChatWebSocketThread.java
laggedHero: I understood that. But without the `@Override` it looks like you create a method just to deprecated it.

app/src/main/java/chat/rocket/android/helper/RxHelper.java
YusukeIwaki: Oh, I didn't know that operator. Thank you.
`Single.defer` works as I expected.

app/src/main/java/chat/rocket/android/service/RocketChatWebSocketThread.java
YusukeIwaki: > But without the @Override it looks like you create a method just to deprecated it.

Hmm, right. I will add `@Override`.

app/src/main/java/chat/rocket/android/helper/RxHelper.java
laggedHero: Can't you just `flatMap` this?

YusukeIwaki: I think an operator like `enumerate()` is needed here.

slice

---

https://github.com/ReactiveX/RxJava/issues/3602

It seems that an operator for getting the index of each item is implemeted in `rxjava-extras`. However it also uses `.zipWith(Observable.range(0, maxRetryCount), (error, retryCount) -> retryCount)`. https://github.com/davidmoten/rxjava-extras/blob/a70df4359a818b240b053cb8bd8d6ce2a56109e9/src/main/java/com/github/davidmoten/rx/util/MapWithIndex.java#L25

I think this is reasonable. However maybe the name of variable `errors` is not good. I will change the variable name.

laggedHero: Yep. Sounds good. 👍

laggedHero:
@YusukeIwaki planning more changes? I'd like to give it a spin with our alpha testers. xD

YusukeIwaki:
@laggedHero No, this is all I wanted to try first.
If some bugs remain, I want fix them with other PRs.
Thank you for your reviewing and good advices!

laggedHero (approved):

PR #178, opened 2017-01-23
Better handling of attachments

### LINKED ISSUE #169: Better message attachment support
Basically #166 #167 and #168 plus more.

The idea is to better support https://rocket.chat/docs/developer-guides/rest-api/chat/postmessage#attachments-detail

laggedHero:
@RocketChat/android

Closes #169 

This is not a final solution since there are some missing points:

- audio attachments
- video attachments
- show the timestamp
- make attachment collapse
- have a thumb for image and opening the full image at a touch
- respect mobile option from settings

Audio and video will need a special solution for each case (with embedded players and so).
The timestamp depends on formatters not available at the module. Instead of copying it or whatever, maybe it would be better to deal with the data better (maybe using MVP).
The collapse behavior may be easy to do. Not sure about that.
Show an image thumb and show the full image is too complex and should have a task of it own.
Respect settings should be a task of it own.

screenshot_1485192648

### DIFF
--- a/app/src/main/java/chat/rocket/android/helper/Avatar.java
+++ b/app/src/main/java/chat/rocket/android/helper/Avatar.java
@@ -30,10 +30,16 @@ public class Avatar {
   };
   private final String hostname;
   private final String username;
+  private final String avatar;
 
   public Avatar(String hostname, String username) {
+    this(hostname, username, null);
+  }
+
+  public Avatar(String hostname, String username, String avatar) {
     this.hostname = hostname;
     this.username = username;
+    this.avatar = avatar;
   }
 
   private static int getColorForUser(String username) {
@@ -76,11 +82,15 @@ public class Avatar {
   private String getImageUrl() {
     //from Rocket.Chat:packages/rocketchat-ui/lib/avatar.coffee
     //REMARK! this is often SVG image! (see: Rocket.Chat:server/startup/avatar.coffee)
-    try {
-      return "https://" + hostname + "/avatar/" + URLEncoder.encode(username, "UTF-8") + ".jpg";
-    } catch (UnsupportedEncodingException exception) {
-      RCLog.e(exception, "failed to get URL for user: %s", username);
-      return null;
+    if (TextUtils.isEmpty(avatar)) {
+      try {
+        return "https://" + hostname + "/avatar/" + URLEncoder.encode(username, "UTF-8") + ".jpg";
+      } catch (UnsupportedEncodingException exception) {
+        RCLog.e(exception, "failed to get URL for user: %s", username);
+        return null;
+      }
+    } else {
+      return avatar;
     }
   }
 
@@ -100,6 +110,10 @@ public class Avatar {
   }
 
   private Drawable getTextDrawable(Context context) {
+    if (username == null) {
+      return null;
+    }
+
     int round = (int) (4 * context.getResources().getDisplayMetrics().density);
 
     return TextDrawable.builder()
@@ -114,16 +128,16 @@ public class Avatar {
 
     // Picasso can be triggered only on Main Thread.
     if (Looper.myLooper() != Looper.getMainLooper()) {
-      new Handler(Looper.getMainLooper()).post(() -> {
-        getBitmap(context, size).continueWith(_task -> {
-          if (_task.isFaulted()) {
-            task.setError(_task.getError());
-          } else {
-            task.setResult(_task.getResult());
-          }
-          return null;
-        });
-      });
+      new Handler(Looper.getMainLooper()).post(() ->
+          getBitmap(context, size)
+              .continueWith(_task -> {
+                if (_task.isFaulted()) {
+                  task.setError(_task.getError());
+                } else {
+                  task.setResult(_task.getResult());
+                }
+                return null;
+              }));
       return task.getTask();
     }
 
--- a/app/src/main/java/chat/rocket/android/helper/Avatar.java
+++ b/app/src/main/java/chat/rocket/android/helper/Avatar.java
@@ -30,16 +30,10 @@ public class Avatar {
   };
   private final String hostname;
   private final String username;
-  private final String avatarUrl;
 
   public Avatar(String hostname, String username) {
-    this(hostname, username, null);
-  }
-
-  public Avatar(String hostname, String username, String avatarUrl) {
     this.hostname = hostname;
     this.username = username;
-    this.avatarUrl = avatarUrl;
   }
 
   private static int getColorForUser(String username) {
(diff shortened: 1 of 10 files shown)


rocket-chat-android-widgets/src/main/res/values/message_styles.xml
```diff
@@ -17,6 +17,9 @@
 
     <style name="TextAppearance.RocketChat.MessageAttachment.Title" parent="TextAppearance.AppCompat.Title">
         <item name="android:textSize">14sp</item>
+    </style>
+
+    <style name="TextAppearance.RocketChat.MessageAttachment.Title.Link" parent="TextAppearance.RocketChat.MessageAttachment.Title">
```
YusukeIwaki: `parent="TextAppearance.RocketChat.MessageAttachment.Title"` is not needed.

The default parent of a style `AAA.BBB.CCC` is `AAA.BBB`.

rocket-chat-android-widgets/src/main/res/layout/message_inline_attachment.xml
```diff
@@ -1,41 +1,109 @@
...
+
+            <android.support.v4.widget.Space
+                    android:layout_width="8dp"
+                    android:layout_height="8dp" />
+
+            <TextView
+                    android:id="@+id/author_name"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:textAppearance="@style/TextAppearance.RocketChat.MessageAttachment.Title.Link"
```
YusukeIwaki: It would be nice if `textAppearance` is set to `@style/TextAppearance.RocketChat.MessageAttachment.Title` and `android:textColor="?android:attr/textColorLink"` manually.

rocket-chat-android-widgets/src/main/res/values/message_styles.xml
YusukeIwaki: If this style will be enhanced in future (ex. bold, big textSize...), the name "Link" would be not proper and another name such as `MessageAttachment.Title.Decorated` or `MessageAttachment.Title.Emphasized` would be better, I think.

YusukeIwaki:
By the way, we would be happy if `https://cdn-demo.rocket.chat/avatar/<username>?_dc=0` has a option to get the image with PNG format. (SVG cannot be parsed easily with Android ImageView... 😓 )

Is there a possibility that such feature is implemented on Rocket.Chat server side?

rocket-chat-android-widgets/src/main/res/layout/message_inline_attachment_field.xml
```diff
@@ -0,0 +1,22 @@
...
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:orientation="vertical">
+
+    <TextView
+            android:id="@+id/field_title"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:textStyle="bold"
+            android:textColor="@android:color/black"
```
YusukeIwaki: It would be better to define a TextAppearance and use it here.

rocket-chat-android-widgets/src/main/res/layout/message_inline_attachment.xml
```diff
@@ -1,41 +1,109 @@
...
+
+            <android.support.v4.widget.Space
+                    android:layout_width="8dp"
+                    android:layout_height="8dp" />
+
+            <TextView
+                    android:id="@+id/text"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    tools:text="Bradley Hilton" />
```
YusukeIwaki, line 92: very private name. Is this OK? 😅

laggedHero: That's the same name used on the attachment documentation. I think it's quite OK. xD

rocket-chat-android-widgets/src/main/res/values/message_styles.xml
laggedHero: The intent was to use this style when you have a link in the title.

So you have the default style and the link style.

app/src/main/java/chat/rocket/android/helper/Avatar.java
```diff
@@ -30,10 +30,16 @@
   };
   private final String hostname;
   private final String username;
+  private final String avatarUrl;
 
   public Avatar(String hostname, String username) {
+    this(hostname, username, null);
+  }
+
+  public Avatar(String hostname, String username, String avatarUrl) {
```
YusukeIwaki: I think `Avatar` class should be designed just for rendering User's avatar, 
and it would be better to define a logic to render Message avatar in MessageRenderer, independent of `Avatar` class.

app/src/main/java/chat/rocket/android/renderer/MessageRenderer.java
```diff
@@ -35,7 +35,7 @@ public MessageRenderer avatarInto(ImageView imageView, String hostname) {
         imageView.setImageResource(R.drawable.ic_error_outline_black_24dp);
         break;
       default:
-        userRenderer.avatarInto(imageView, hostname);
+        userRenderer.avatarInto(imageView, hostname, object.getAvatar());
```
YusukeIwaki: I think UserRenderer should not know Message's avatar.
It would be better to check the presence of Message's avatar here, and request to UserRenderer to render avatar image only when Message's avatar is absent.

```
if (object.getAvatar != null) {
  loadImageInto(object.getAvatar(), imageView);  // just load image with Picasso.
} else {
  userRenderer.avatarInto(imageView, hostname);
}
```

app/src/main/java/chat/rocket/android/renderer/MessageRenderer.java
```diff
@@ -45,7 +45,7 @@ public MessageRenderer avatarInto(ImageView imageView, String hostname) {
    * show Username in textView.
    */
   public MessageRenderer usernameInto(TextView textView) {
-    userRenderer.usernameInto(textView);
+    userRenderer.usernameInto(textView, object.getAlias());
```
YusukeIwaki: like avatar, it would be better to set alias text here, independent of UserRenderer.

```
if (!TextUtils.isEmpty(object.getAlias)) {
  setAliasTextInto(object.getAlias(), object.getUser.getName(), textView);
} else {
  userRenderer.usernameInto(textView);
}
```

laggedHero:
About SVG: maybe we may find something to deal with that. Fresco has support for custom decoders (https://github.com/facebook/fresco/issues/1008) and that may help.

app/src/main/java/chat/rocket/android/renderer/MessageRenderer.java
laggedHero: I'll open (make public) the `getTextDrawable` from `Avatar` (it's the main reason to delegate any rendering to it).

app/src/main/java/chat/rocket/android/renderer/MessageRenderer.java
laggedHero: `object.getUser()` is null. Quite not useful. I'll try and see what happens.

YusukeIwaki:
> About SVG: maybe we may find something to deal with that. Fresco has support for custom decoders (facebook/fresco#1008) and that may help.

Thank you for the good news!

YusukeIwaki:
[Material design guideline](https://material.io/guidelines/style/color.html#color-color-schemes) says that pure black text color should be used only for important text.
So actually it would be better to set `alpha=87%` for some TextViews.

However almost great 👍  Let's modify them after merging this PR.


I really appreciate for your contribution!

YusukeIwaki (approved):

PR #180, opened 2017-01-25
improve the logic to render username and message alias

YusukeIwaki:
@RocketChat/android

Originally, MessageRenderer and UserRenderer set both text contents and text styles.
However for example, if a TextView is put on the dark background, the textColor should be white instead of black.
So, it would be better that render just put text contents into TextView, and style is set in layout xml.

This PR adds "sub_username" TextView and changes the logic for rendering alias and username as below:

image

### DIFF
--- a/app/src/main/res/values/message_styles.xml
+++ b/app/src/main/res/values/message_styles.xml
@@ -2,4 +2,10 @@
 <resources>
     <color name="newday_color">#e0e0e0</color>
     <color name="newday_text_color">#444444</color>
+    <style name="TextAppearance.RocketChat.Message.Username" parent="TextAppearance.AppCompat.Body2">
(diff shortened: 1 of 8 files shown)


app/src/main/res/values/message_styles.xml
```diff
@@ -2,4 +2,10 @@
 <resources>
     <color name="newday_color">#e0e0e0</color>
     <color name="newday_text_color">#444444</color>
+    <style name="TextAppearance.RocketChat.Message.Username" parent="TextAppearance.AppCompat.Body2">
+        <item name="android:textColor">#bb000000</item>
```
YusukeIwaki: I will define `textColorPrimary` and `textColorSecondary` in theme.

laggedHero (approved):

PR #188, opened 2017-01-27
Using Fresco to render images

### LINKED ISSUE #181: GIF support
The web version of Rocket.Chat will render GIFs as part of its link rendering. The mobile app will not.

laggedHero:
@RocketChat/android

Closes #181 

This is the initial work on Fresco support.


laggedHero:
No more picasso. Loads PNG, JPG, WEBP, GIF and SVG. :)

Bug reports to follow. xD

YusukeIwaki:
Awesome work. GIF is rendered a bit more beautifully than Glide's!

rocket-chat-android-widgets/src/main/res/layout/message_inline_attachment.xml
```diff
@@ -93,14 +96,12 @@
...
         </LinearLayout>
 
-        <ImageView
+        <com.facebook.drawee.view.SimpleDraweeView
                 android:id="@+id/image"
-                android:layout_width="wrap_content"
-                android:layout_height="wrap_content"
-                android:maxHeight="200dp"
+                android:layout_width="match_parent"
+                android:layout_height="200dp"
```
YusukeIwaki: slice

It seems that the boundary of each attachment image is larger than expected for its size.


Fresco doesn't support `wrap_content` ...
http://frescolib.org/docs/wrap-content.html#_
However a developer in facebook provides a [workaround](http://stackoverflow.com/questions/33955510/facebook-fresco-using-wrap-conent) for this. Can we use this...?

YusukeIwaki: The image is unexpectedly collapsed with setting `layout_height=wrap_content` ...
Hmm...

laggedHero: We could add some code to resize the view after we have the image. There's some example code of it done by a Fresco dev.

But they don't recommend that. Anyway it could be a temporary fix until the server gives us the image size.

YusukeIwaki: Hmm, I see.

app/src/main/java/chat/rocket/android/api/rest/CookieInterceptor.java
```diff
@@ -0,0 +1,27 @@
...
+
+  private final CookieProvider cookieProvider;
+
+  public CookieInterceptor(CookieProvider cookieProvider) {
+    this.cookieProvider = cookieProvider;
+  }
+
+  @Override
+  public Response intercept(Chain chain) throws IOException {
+    if (chain.request().url().host().equals(cookieProvider.getHostname())) {
```
YusukeIwaki, line 19: This logic probably doesn't work when multiple server configuration exists (in future).
Some kind of iteration for ` [host] => [cookie]` would be required here.

rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/message/RocketChatAvatar.java
```diff
@@ -0,0 +1,66 @@
+package chat.rocket.android.widget.message;
```
YusukeIwaki: This class is not only for message (used also in drawer menu). So it would be better to put this class in `chat.rocket.android.widget`

rocket-chat-android-widgets/src/main/res/layout/message_inline_attachment.xml
laggedHero: I tried it here: The scroll is bad.. really bad. We could improve that with an initial smaller size. Will try that.

Tha SVG? Won't show. We'll need some special code to deal with it.

app/src/main/java/chat/rocket/android/api/rest/CookieInterceptor.java
laggedHero: Are we rendering images from a host while "looking" at other?

rocket-chat-android-widgets/src/main/res/layout/message_inline_attachment.xml
```diff
@@ -93,14 +96,12 @@
...
                 android:layout_marginBottom="8dp"
-                android:adjustViewBounds="true"
-                android:scaleType="fitStart" />
+                fresco:actualImageScaleType="fitCenter" />
```
YusukeIwaki: I think `fitStart` is better here. 

Every image is put on the left-side in web-client
image

rocket-chat-android-widgets/src/main/res/layout/message_inline_attachment.xml
laggedHero: The scroll is bad. For real.
I'll make the view smaller a bit. Maybe it will help

rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/RocketChatWidgets.java
```diff
@@ -0,0 +1,26 @@
...
+import android.content.Context;
+import com.facebook.drawee.backends.pipeline.DraweeConfig;
+import com.facebook.drawee.backends.pipeline.Fresco;
+import com.facebook.imagepipeline.backends.okhttp3.OkHttpImagePipelineConfigFactory;
+import com.facebook.imagepipeline.core.ImagePipelineConfig;
+
+import chat.rocket.android.widget.fresco.ImageFormatConfigurator;
+import okhttp3.OkHttpClient;
+
+public class RocketChatWidgets {
```
YusukeIwaki, line 13: I think it is better for this class to define as fresco-specific, such as `chat.rocket.android.widget.fresco.FrescoConfiguration`

When another initialization logic is added in future, it would be better to define another class if it is not related to fresco.

rocket-chat-android-widgets/src/main/res/layout/message_inline_attachment.xml
laggedHero: I removed the resize feat. No good at all for the scrolling.

rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/RocketChatWidgets.java
laggedHero: From an external caller (`app` module) what we are initializing is the `RocketChatWidgets` module, not a specific library inside it.

app/src/main/java/chat/rocket/android/api/rest/CookieInterceptor.java
laggedHero: I don't think this is a problem, since we'll not be rendering images from two servers at the same time. That's my point.

YusukeIwaki: I got it. You're right. :)

rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/RocketChatWidgets.java
YusukeIwaki: Yes, it's right that caller shouldn't know the implementation of image renderer (fresco).

However I think it would be better for classname to imply when to initialize or why the parameter of OkHttpClient is needed, even just a little...

YusukeIwaki: Hmm, I've changed my mind a little... Leave this classname as it is, for now.

Also, https://github.com/RocketChat/Rocket.Chat.Android/pull/188/commits/32aceba380c2c5b9de1ed1f86bb481fa3dfceed7 looks better :)

Sorry for confusing, and thank you for discussion.

laggedHero: Docs would be enough for that.

app/src/main/java/chat/rocket/android/api/rest/CookieInterceptor.java
```diff
@@ -0,0 +1,27 @@
...
+      Request newRequest = chain.request().newBuilder()
+          .header("Cookie", cookieProvider.getCookie())
```
YusukeIwaki: I think it is better not to add the "Cookie" header if it is empty string.

YusukeIwaki (approved):
Awesome! 👍
Thank you for providing such a great feature!

Please check https://github.com/RocketChat/Rocket.Chat.Android/pull/188#discussion_r98250482 before merging.

rocket-chat-android-widgets/src/main/res/layout/message_inline_attachment.xml
YusukeIwaki: Thank you for your effort and trial.
Let's solve this after server-side changes.

laggedHero:
CircleCI didn't phone home with the success result. I'll just merge this anyway.

Issue #184, opened 2017-01-26
Menu blank


geekgonecrazy:
Your Rocket.Chat.Android version: beta6

![screenshot_20170125-230848](https://cloud.githubusercontent.com/assets/51996/22320611/5b259b88-e353-11e6-997d-032e3fd654f4.png)

Seems to happen at some point after being backgrounded.

Swiping away does not work.  Force stopping doesn't work.  Force Stop + Clear Cache doesn't work.

Only thing that seems to fix it is wiping data

geekgonecrazy:
Seems you get it in a state where the websocket is hanging from being in the background a while.  In my case I could tell websocket must not be connected because I had a notification for a DM that wasn't in my list.  So I swiped the card away.  When I re-opened app it was broken like this again :(

Probably 6-7th time so far

laggedHero:
I've found one possibility for this bug to happen and it's related to push notifications.

I have a PR open for that.

And yes: only by wiping data things go back to normal. The PR should helps with that too.

geekgonecrazy:
Testing that PR, will report back :+1:

laggedHero:
#191 may have fixed this. Let's close this for now.

PR #191, opened 2017-01-30
Push notifications was breaking things for me

laggedHero:
@RocketChat/android


YusukeIwaki (approved):
Thank you!

geekgonecrazy:
Tested most of today seems to fix #184

laggedHero:
@geekgonecrazy I'll close #184 for now then.

Let's hope it stays like that. xD

PR #193, opened 2017-01-30
Fix incorrect behavior

### LINKED ISSUE #185: Message sent but UI says not sent
Your Rocket.Chat.Android version: beta6

screenshot_20170125-235319

As they say a picture is worth a thousand words... :)

laggedHero:
@RocketChat/android

Closes #185 

We're setting the message sync state in several places and that generates a racing condition. Sometimes the correct result would show up.

The problem lies with the success response for sending a message: it has an empty message.

I added a special clause to that and removed the sync setter from `NewMessageObserver`. We may rely the one from `stream-room-messages`.

### DIFF
--- a/app/src/main/java/chat/rocket/android/api/MethodCallHelper.java
+++ b/app/src/main/java/chat/rocket/android/api/MethodCallHelper.java
@@ -317,8 +317,17 @@ public class MethodCallHelper {
    */
   private Task<JSONObject> sendMessage(final JSONObject messageJson) {
     return call("sendMessage", TIMEOUT_MS, () -> new JSONArray().put(messageJson))
-        .onSuccessTask(CONVERT_TO_JSON_OBJECT)
-        .onSuccessTask(task -> Task.forResult(Message.customizeJson(task.getResult())));
+        .onSuccessTask(task -> {
+          final String result = task.getResult();
+          if (result.equals("")) {
+            // valid success result
+            return Task.forResult(new JSONObject());
+          }
+
+          // should validate any other response
+          final JSONObject jsonObject = new JSONObject(result);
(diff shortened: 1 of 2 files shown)


app/src/main/java/chat/rocket/android/api/MethodCallHelper.java
```diff
@@ -317,8 +317,17 @@ public MethodCallHelper(RealmHelper realmHelper, DDPClientRef ddpClientRef) {
    */
   private Task<JSONObject> sendMessage(final JSONObject messageJson) {
     return call("sendMessage", TIMEOUT_MS, () -> new JSONArray().put(messageJson))
-        .onSuccessTask(CONVERT_TO_JSON_OBJECT)
-        .onSuccessTask(task -> Task.forResult(Message.customizeJson(task.getResult())));
+        .onSuccessTask(task -> {
+          final String result = task.getResult();
+          if (result.equals("")) {
+            // valid success result
+            return Task.forResult(new JSONObject());
```
YusukeIwaki: `{_id: <messageId>, syncstate: SYNCED}` is not needed here?

YusukeIwaki:
Thank you for your proposal.

> We may rely the one from stream-room-messages.

This direction is very good, and I completely agree with it.

However if so, I think it is better to change `Task<JSONObject> sendMessage` to `Task<Void> sendMessage` as `sendFileMessage` already changed in [this commit](https://github.com/RocketChat/Rocket.Chat.Android/commit/d0cfaf649bc983b9288e46a18dfac155d90cc83e) . It would clarify that `MethodCallHelper#sendMessage` is unconcerned with the returned `Message` object.

YusukeIwaki (approved):
👍  Thank you!

PR #198, opened 2017-02-08
Un-Realm-app the app

laggedHero:
Right now Realm is the backbone and soul of the app. And that's no good.

We need to be able to use Realm as a solution for persistence that may be exchangeable if needed. It is especially important if we think about providing an SDK in the future based on the work done here.

In order to do it, I'm using `auto-value` for our entities definition and ~ab~using  some patterns (MVP, Repository).

More (a lot more) changes to come.


laggedHero:
I'm thinking about stopping for now. We do have the need for new features and so on.

It's possible that I had expressed my intent already. xD

I'll continue to improve things little by little. Interactors should become so (right now I'm being somewhat creative with them) and there's still a lot of hard realm dependencies.

We should look for some way to deal with dependency creation and injection. `dagger2` looks nice, but we do have options (just need to look out for them).

Anyway.

`rocket-chat-core` is where our business rules should live.
`rocket-chat-android-widgets` are our widgets.
`persistence-realm` will be a reference persistence implementation based on `Realm`. It will implement any `Repository` defined by the `core` and deal with converting objects.
`android-ddp` should be the ddp layer. We'll need to interface it with `core` later on.
`app` will depend on those guys. It will wire them together and be an app.

laggedHero:
Oh, yeah. @YusukeIwaki take a look. :)

persistence-realm/build.gradle
```diff
@@ -39,6 +41,9 @@ android {
 dependencies {
   testCompile 'junit:junit:4.12'
   compile project(':log-wrapper')
+  compile project(':rocket-chat-core')
+  compile 'io.reactivex:rxjava:1.2.3'
+  compile 'io.reactivex:rxandroid:1.2.1'
```
YusukeIwaki, line 46: Please use `roomProject.ext.rxJava` instead for frequently used libs.

app/src/main/java/chat/rocket/android/activity/LoginContract.java
```diff
@@ -0,0 +1,17 @@
+package chat.rocket.android.activity;
+
+import chat.rocket.android.shared.BaseContract;
+
+public interface LoginContract {
+
+  interface View extends BaseContract.View {
```
YusukeIwaki: The name `View` is a little confusing. It is nothing to do with `android.view.View`.

`Behavior` or `Action`  would be better.

app/src/main/java/chat/rocket/android/activity/LoginContract.java
```diff
@@ -0,0 +1,17 @@
...
+    void showLogin(String hostname);
+
+    void showRetryLogin(String hostname);
+
+    void close();
```
YusukeIwaki: `close()` is not good. 


For example: 

```
    if (hostname == null || hostname.length() == 0) {
      view.close();
      return;
    }
```

It is hard to guess what `view.close()` does.

Please use another name like `hideLogin()`

YusukeIwaki:
@laggedHero Thank you for your hard development :)

Actually the amount of this codes is very huge, and hard to review... 😓 
So I believe your skill for now!

This is just my opinion, but currently (before this PR) many objects has the responsibility for building Realm query. This is the most important problem, I think.
It seems that Rocket.Chat.iOS has the classes `XXXManager` or `XXXHandler` for handling the Realm query and other classes doesn't touch Realm directly.
So I think it would be better if such wrapper classes exist as Rocket.Chat.iOS for making the Realm queries more human-readable.

persistence-realm/build.gradle
laggedHero: To tell you the truth: I'll undo all `roomProject.ext.rxJava` stuff in the project. I'd like to see the modules as independent as possible from each other.

app/src/main/java/chat/rocket/android/activity/LoginContract.java
laggedHero: `closeView` so. Cause that's what we do.

laggedHero:
I'll leave any `Realm` stuff on the specific `Realm` repository. So only the repository will know the `Realm` specifics and the `interactors` will have the rules for messaging.

But let me update that method first.

PR #201, opened 2017-02-12
Use external library for FontAwesome widgets

YusukeIwaki:
@RocketChat/android

Just a refactoring.
I created a library for providing FontAwesomeTextView and FontAwesomeButon with `@string/fa_xxx` definitions. (because I use them in 3 hobby projects... 😃  )
Shall we use the library also in Rocket.Chat.Android?


app/build.gradle
```diff
@@ -82,13 +82,6 @@ play {
     track = "${track}"
 }
 
-repositories {
-    mavenCentral()
-    maven { url 'https://github.com/lijingle1/stetho-realm/raw/master/maven-repo' }
-    maven { url 'http://dl.bintray.com/amulyakhare/maven' }
-    maven { url "https://clojars.org/repo/" } //for icepick.
```
YusukeIwaki, line 89: Moving these definition is not directly related to introducing widget-fontawesome. 
However, I think it would be better to define these in one place (root build.gradle).

laggedHero (approved):
Looks good

PR #260, opened 2017-04-23
fix lost text when turn off then turn on screen

### LINKED ISSUE #256: [BUG] Text in input lost when screen sleeps
Your Rocket.Chat.Android version: latest

I've had several times where I have a big message typed and someone starts talking to me in person.  So I stop typing.  Screen goes to sleep.  Then I wake it and message is gone. 

So right now if its important I have to keep tapping the phone to keep it awake until I can hit send :grin:

congnt24:
fix lost text when turn off then turn on screen, hide keyboard when change room chat

@RocketChat/android

Closes #256

### DIFF
--- a/app/src/main/java/chat/rocket/android/activity/AbstractAuthedActivity.java
+++ b/app/src/main/java/chat/rocket/android/activity/AbstractAuthedActivity.java
@@ -36,7 +36,6 @@ abstract class AbstractAuthedActivity extends AbstractFragmentActivity {
     if (savedInstanceState == null) {
       handleIntent(getIntent());
     }
-    ConnectivityManager.getInstance(getApplicationContext()).keepAliveServer();
 
     updateHostnameIfNeeded(rocketChatCache.getSelectedServerHostname());
     updateRoomIdIfNeeded(rocketChatCache.getSelectedRoomId());
@@ -154,6 +153,8 @@ abstract class AbstractAuthedActivity extends AbstractFragmentActivity {
     super.onResume();
 
     subscribeToConfigChanges();
+
+    ConnectivityManager.getInstance(getApplicationContext()).keepAliveServer();
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


app/src/main/java/chat/rocket/android/activity/MainActivity.java
```diff
@@ -51,7 +52,7 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
   @Override
   protected void onPause() {
     if (presenter != null) {
-      presenter.release();
+//      presenter.release();
```
laggedHero: This is really bad since we'll not clean up the subscriptions made. I can't accept this change.

app/src/main/java/chat/rocket/android/activity/AbstractAuthedActivity.java
```diff
@@ -36,6 +36,10 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
     if (savedInstanceState == null) {
       handleIntent(getIntent());
     }
+    ConnectivityManager.getInstance(getApplicationContext()).keepAliveServer();
```
laggedHero: Let's put this back into the `onResume`. It's trying to keep the connection alive.

laggedHero (changes requested):

congnt24:
The problem is that whenever onResume() call, it will recreate a new RoomFragment and the state of fragment will lose, not just text in EditText. I already changed some code that you requested. Please review.

app/src/main/java/chat/rocket/android/activity/MainActivity.java
```diff
@@ -172,6 +181,8 @@ public void showHome() {
   public void showRoom(String hostname, String roomId) {
     showFragment(RoomFragment.create(hostname, roomId));
     closeSidebarIfNeeded();
+    //TODO: Hide keyboard
+    KeyboardHelper.hideSoftKeyboard(this);
```
geekgonecrazy: Looks like might have a comment that needs removed.  The line right below seems to be doing exactly what the TODO calls for :)

laggedHero:
There's still some rough edges... but the code here, in general, is kind of rough.

Let's remove that comment and merge it in.

congnt24:
I've removed comment

laggedHero (approved):

laggedHero:
Hmm, I'll check the build here.

Issue #279, opened 2017-05-07
Image avatars are not round


gardenerik:
Your Rocket.Chat.Android version: latest

If user has avatar (and if it loads properly - #251), it displays as square, not rounded, as the default inicial avatars are.

PR #283, opened 2017-05-11
Fix notification and round avatar

### LINKED ISSUE #280: Open the correct from when tapping in a notification
Your Rocket.Chat.Android version: 1.0.11
When I'm in GENERAL room, then i press Back to let app in background (user1).
I use another account (user2) and send a private message to user1.
Notification appear in device of user1, I open rocketchat app (still in GENERAL room) then click notification but nothing happend. 
->It should open the chat room with user2

congnt24:
@RocketChat/android

Closes #280 #279 

When a user in the app, then they click notification. Activity will call onPause, then onNewIntent then onResume. But when onPause is called, the Subscription to rocketChatCache will be released and when onNewIntent call, the change of hostname, roomid will not be handle. So It won't open new room.
I've fixed it.

### DIFF
--- a/rocket-chat-android-widgets/src/main/res/layout/message_avatar.xml
+++ b/rocket-chat-android-widgets/src/main/res/layout/message_avatar.xml
@@ -4,4 +4,5 @@
         android:id="@+id/drawee_avatar"
         android:layout_width="match_parent"
         android:layout_height="match_parent"
-        fresco:actualImageScaleType="fitCenter" />
+        fresco:actualImageScaleType="fitCenter"
+        fresco:roundedCornerRadius="5dp" />
--- a/app/src/main/java/chat/rocket/android/activity/AbstractAuthedActivity.java
+++ b/app/src/main/java/chat/rocket/android/activity/AbstractAuthedActivity.java
@@ -4,21 +4,21 @@ import android.content.Intent;
 import android.os.Bundle;
 import android.support.annotation.Nullable;
 
-import io.reactivex.android.schedulers.AndroidSchedulers;
-import io.reactivex.disposables.CompositeDisposable;
-import io.reactivex.schedulers.Schedulers;
-
 import java.util.List;
+
 import chat.rocket.android.LaunchUtil;
 import chat.rocket.android.RocketChatCache;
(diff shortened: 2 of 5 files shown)


rocket-chat-android-widgets/src/main/res/layout/message_avatar.xml
```diff
@@ -4,4 +4,5 @@
         android:id="@+id/drawee_avatar"
         android:layout_width="match_parent"
         android:layout_height="match_parent"
-        fresco:actualImageScaleType="fitCenter" />
\ No newline at end of file
+        fresco:actualImageScaleType="fitCenter"
+        fresco:roundedCornerRadius="5dp" />
```
rafaelks, line 8: Can't we have this `value` for all avatars in the app in a more generic way?

congnt24: It's generic because it's in a custom view. You can use it somewhere without adding/changing that value

rafaelks: Will this value be used in the left-side menu too? By generic, I was talking about all avatars in the app.

congnt24: Yes, the avatar in the left-side also use that value.

rafaelks: Oh, sounds good! I was thinking it would not use because of the name `message_avatar`. Thank you! 👍

rafaelks:
@congnt24 Do you mind signing our CLA? It's required to merge your PR: https://cla-assistant.io/RocketChat/Rocket.Chat.Android?pullRequest=283

congnt24:
@rafaelks I've signed the CLA for RocketChat/Rocket.Chat.Android.

rafaelks:
@congnt24 Thank you!

matheusjardimb (approved):
looks good to me!

matheusjardimb:
@rafaelks would you trigger a "Rebuild without cache" into CircleCI? 
I don't have enough permissions for that yet and it doesn't seems to have problems with the PR

Issue #268, opened 2017-04-27
Rocket.Chat+ on Android: Sending message stuck in "Sending" state, displaying wrong message owner


somsak:
Rocket.Chat Version: 0.55.1
Running Instances: 1
DB Replicaset OpLog: 0
Node Version: 4.8.1 (We use Official Docker image)
Android app version: 1.0.10

When sending message from Android app. Most of the time it will stuck in the "Sending" state and showing wrong owner of the message (It should be the owner of the phone, but it randomly pick someone around and show the message state as Sending).

This seems to happen when we exit the app for a while and comeback (Usually from push notification). The App works fine the first time we login to the server.

When the message was in Sending state, the message was actually already delivered (by checking on Web-site or Cordova app).

Sometime the message was stuck in "!" state and need resend, although the message was actually delivered.

The current workaround is to push "Back" button to exit the app, or kill the app and re-open it.

This didn't happen with Cordova app.

NOTE: We use Rocket.chat Gateway in push notification setting.

somsak:
Today I can confirmed that this problem also occurred with https://demo.rocket.chat

Here's an example screenshot of what happened. It stuck in "Sending" state for a long time.

![smartselectimage_2017-05-03-16-19-05](https://cloud.githubusercontent.com/assets/353565/25671146/27067018-3059-11e7-9b51-57b552a73f0f.png)

adam-kulagowski:
This issue also affect our team/company at codilime.rocket.chat - it kinda blocking point for native app.

djschilling:
We also have this problem with our instance https://chat.ct-srv.de.

guptamonu4:
I am also facing same problem in Rocket chat+.

rafaelks:
This is very important, will add the urgent label to it.

TheReal1604:
@rafaelks Can also confirm this.

rafaelks:
This is in progress already.

joneswe:
the wrong message owner getting displayed, because the recyclerView doesn't overwrite the views of messages which are not synced. Caused by missing attributes.

The interesting Question is: Why do we get Problems by sending and receiving messages after an instance of time or after we turn off and on the internet connection

matheusjardimb:
Hello there! Let me help on this one!

I see clearly two different problems listed in this issue: 
- the first one is related to the recyclerView, as @joneswe said. This part is duplicated of #251 and should not be happening anymore.
[...]

PR #332, opened 2017-07-01
Bumped timeout from 4000 to 20000 ms as it seems a more reasonable

leonardoaramaki:
@RocketChat/android

Bumped timeout for rpc calls from 4000 to 20000 millis as it seems a more reasonable threshold.
    
Also updating a sending message sync state to SyncState.SYNCED at Realm in case a sending message is successfully sent.

This sync status update fix is related to issue #268


rafaelks:
Thanks @leonardoaramaki! LGTM, can you review the code @matheusjardimb, please?

matheusjardimb (approved):

PR #334, opened 2017-07-01
Fix #268

### LINKED ISSUE #268: Rocket.Chat+ on Android: Sending message stuck in "Sending" state, displaying wrong message owner

philipbrito:
@RocketChat/android

Closes #268


rafaelks:
LGTM! Can you review the change @matheusjardimb, please?

app/src/main/java/chat/rocket/android/renderer/MessageRenderer.java
```diff
@@ -76,10 +75,11 @@ public MessageRenderer timestampInto(TextView textView) {
     }
 
     switch (object.getSyncState()) {
```
leonardoaramaki, line 77: Shouldn't it be better if we put this way:

case SyncState.FAILED:
         textView.setText(R.string.failed_to_sync);
         break;

philipbrito: Nice catch! But, what is the diference between FAILED and NOT_SYNCED state?
@rafaelks ?

leonardoaramaki: @filipedelimabrito As I understand so far, NOT_SYNCED is the first state of a message before it's sent, when it's first created.

matheusjardimb: @leonardoaramaki I agree! `NOT_SYNCED` seems to be the first state. `MessageInteractor::send` exemplifies that. The state then becomes `SYNCING` and later either `SYNCED` or `FAILED`.

rafaelks:
@RocketChat/android is it good to merge?

leonardoaramaki (approved):

PR #339, opened 2017-07-04
Add waiting view to hostname screen/layout

philipbrito:
Like the Login screen does, this PR adds the waiting view to hostname screen, thus, the user will be able to see that the connection is in progress.

### DIFF
--- a/app/src/main/res/layout/fragment_input_hostname.xml
+++ b/app/src/main/res/layout/fragment_input_hostname.xml
@@ -1,33 +1,41 @@
 <?xml version="1.0" encoding="utf-8"?>
 <FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
-        xmlns:app="http://schemas.android.com/apk/res-auto"
-        xmlns:tools="http://schemas.android.com/tools"
-        android:layout_width="match_parent"
-        android:layout_height="match_parent"
-        android:background="?attr/colorPrimaryDark">
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:background="?attr/colorPrimaryDark"
+    tools:context="chat.rocket.android.fragment.add_server.InputHostnameFragment">
 
     <LinearLayout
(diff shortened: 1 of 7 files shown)


app/src/main/res/layout/fragment_input_hostname.xml
```diff
@@ -1,63 +1,79 @@
...
                 android:layout_width="0px"
                 android:layout_height="wrap_content"
                 android:layout_weight="1"
                 android:orientation="vertical">
 
-            <TextView
+                <TextView
                     android:layout_width="match_parent"
                     android:layout_height="wrap_content"
+                    android:paddingStart="4dp"
```
matheusjardimb, line 33: should this become move to @dimen/padding_4, similar to @dimen/margin_8 used in this very same file?

matheusjardimb: not sure if that make sense :)

matheusjardimb:
Isn't it weird that most layout (xml) files use mixed tab and spaces as indentation? 
It becomes really annoying when reviewing code changes ... :disappointed:

app/src/main/res/layout/fragment_input_hostname.xml
philipbrito: It is a workaround to align the TextView w/ the EditText. Take a look:

Before
before

After:
after

philipbrito:
Should we use [intellij-java-google-style.xml](https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml) @matheusjardimb ? This [file](https://github.com/RocketChat/Rocket.Chat.Android/blob/develop/config/quality/style-guide/intellij-google-style.xml) seems to refer to it, but is invalid.

matheusjardimb:
Sure, I guess that should help removing the noise indentation can generate. 
In this case, I'd be great to have some instructions in the README file in order to help new contributors use this style. Know what I'm sayin?

philipbrito:
I agree! I'll do that!

app/src/main/res/layout/fragment_input_hostname.xml
matheusjardimb: I see. Not big deal tho

matheusjardimb (approved):

PR #346, opened 2017-07-11
RecyclerView wrong avatar solution

philipbrito:
@RocketChat/android

This is a WIP PR to solve the wrong avatar which sometimes is displayed on a chat room.
To see this bug follow theses steps:

1- Open the RC application w/ an active internet connection.
2- Disconnect the internet connection on a chat room.
3- Write some message and send it.
4- Now, scroll up and down to see the wrong avatar which is shown to your most recently message (its status should be equals FAILED to send).


philipbrito:
I think that at this point the last commits solve the problem discussed [above](https://github.com/RocketChat/Rocket.Chat.Android/pull/346#issue-242028575) (steps from 1 to 4). Now, lets do more/another tests....

rafaelks (approved):

PR #354, opened 2017-07-15
Fix/websocket drops keeping bolts

leonardoaramaki:
@RocketChat/android

WIP: Did some work on the network layer without entirely rewriting it, that being the fastest solution as possible on short time. These changes are supposed to handle websocket connection drops due to network intermittences such as changing from wi-fi to mobile network or total internet connection drops. The present PR is not supposed to be a permanent solution altogether and probably does not handle all edge cases.


android-ddp/src/main/java/chat/rocket/android_ddp/DDPClientImpl.java
```diff
@@ -369,6 +378,28 @@ public void unsubscribeBaseListeners() {
...
   }
 
+  public Task<RxWebSocketCallback.Failure> getOnFailureCallback() {
+    TaskCompletionSource<RxWebSocketCallback.Failure> task = new TaskCompletionSource<>();
+
+    flowable.filter(callback -> callback instanceof RxWebSocketCallback.Failure)
+            .cast(RxWebSocketCallback.Failure.class)
+            .subscribe(
+                    task::setResult,
+                    err -> {
```
luciofm: When `onSuccess` or `onError` have more than one line, extract it to a private method with a meaningful name, increases readability and make clear for other devs the intention.
Ex
```
.subscribe(task::setResult, this::setTaskError(task, throwable));

private void setTaskErro(TaskCompletionSource<RxWebSocketCallback.Failure> task, Throwable throwable) {
    if (err instanceof Exception) {
        task.setError((Exception) err);
    } else {
        task.setError(new Exception(err));
    }
}
```

android-ddp/src/main/java/chat/rocket/android_ddp/DDPClientImpl.java
```diff
@@ -383,7 +414,8 @@ private boolean sendMessage(String msg, @Nullable JSONBuilder json) {
   private void sendMessage(String msg, @Nullable JSONBuilder json,
                            TaskCompletionSource<?> taskForSetError) {
     if (!sendMessage(msg, json)) {
-      taskForSetError.trySetError(new DDPClientCallback.Closed(client));
+      if (taskForSetError != null)
```
luciofm: Disclaimer: In the next few days I'll be creating an Coding Style/Guideline doc, and a review guidelines to keep and improve our code quality

Single `if` statements must have braces `{ }` or have the body in the same line.


```
// preferred style
if (condition) {
    body();
}

// also acceptable if fits right in one line
if (condition) body();

// not acceptable
if (condition)
    body();
```

app/src/main/java/chat/rocket/android/activity/MainPresenter.java
```diff
@@ -3,22 +3,24 @@
 import android.support.annotation.NonNull;
 import android.support.v4.util.Pair;
 
-import io.reactivex.Flowable;
```
luciofm, line 6: I'll be creating a default Android Studio config, so we can avoid this import reordering on PRs.

app/src/main/java/chat/rocket/android/activity/MainPresenter.java
```diff
@@ -159,6 +161,19 @@ private void subscribeToSession() {
...
+    addSubscription(
+            RxJavaInterop.toV2Observable(connectivityManagerApi.getServerConnectivityAsObservable())
+                    .subscribeOn(Schedulers.io())
+                    .observeOn(AndroidSchedulers.mainThread())
+                    .subscribe(serverConnectivity -> {
+                      if (serverConnectivity.state == ServerConnectivity.STATE_CONNECTING) {
+                        view.showConnecting();
+                      }
+                    },
+                    err -> {
```
luciofm: Don't ignore exceptions, do at least a `err.printStackTrace();` or if it is really safe to ignore it, add a comment explaining why.

Ignoring exceptions can make it harder to understand why something is not working, and printing the stacktrace make it obvious where the problem is,

app/src/main/java/chat/rocket/android/activity/MainPresenter.java
```diff
@@ -159,6 +161,19 @@ private void subscribeToSession() {
         );
 
     addSubscription(subscription);
+
+    addSubscription(
+            RxJavaInterop.toV2Observable(connectivityManagerApi.getServerConnectivityAsObservable())
```
luciofm: Could you please create an Issue to update the code base to RxJava 2, and add a TODO comment pointing to the issue whenever we need to use the RxJavaInterop for our own code.

app/src/main/java/chat/rocket/android/service/RocketChatWebSocketThread.java
```diff
@@ -3,24 +3,24 @@
...
 import java.util.Iterator;
+
 import bolts.Task;
+import chat.rocket.android.RocketChatCache;
 import chat.rocket.android.api.DDPClientWrapper;
 import chat.rocket.android.api.MethodCallHelper;
 import chat.rocket.android.helper.LogIfError;
 import chat.rocket.android.helper.TextUtils;
 import chat.rocket.android.log.RCLog;
-import chat.rocket.core.models.ServerInfo;
```
luciofm: We should try to avoid reordering imports on commits that also adds or removes imports, it makes harder to see what actually changed.

app/src/main/java/chat/rocket/android/service/RocketChatWebSocketThread.java
```diff
@@ -238,6 +252,35 @@ public void run() {
                   return null;
                 });
 
+                task.getResult().client.getOnFailureCallback().onSuccess(_task -> {
+                  ddpClient = null;
+                  CompositeDisposable subscriptions = new CompositeDisposable();
```
luciofm: Try to keep a consistent naming, these are disposables not subscriptions :smiley:

app/src/main/java/chat/rocket/android/service/RocketChatWebSocketThread.java
```diff
@@ -238,6 +252,35 @@ public void run() {
...
+                  connectivityManager.notifyConnecting(hostname);
+                  subscriptions.add(
+                          RxJavaInterop.toV2Single(connect().retry())
+                                  .subscribe(
```
luciofm: Extract to a method with meaningful name.

app/src/main/java/chat/rocket/android/service/RocketChatWebSocketThread.java
```diff
@@ -238,6 +252,35 @@ public void run() {
...
+                                                      realmHelper,
+                                                      ddpClientRef,
+                                                      roomId
+                                              );
+                                              streamRoomObserver.register();
+                                              listeners.add(streamRoomObserver);
+                                            }
+                                          },
+                                          err -> {
+                                            subscriptions.dispose();
```
luciofm: Don't ignore the exception. Is it safe to assume that no more subscriptions will be added to the CompositeDisposable after this?

app/src/main/java/chat/rocket/android/service/observer/MethodCallObserver.java
```diff
@@ -117,10 +118,10 @@ public void onUpdateResults(List<MethodCall> results) {
           } else if (exception instanceof DDPClientCallback.RPC.Timeout) {
             // temp "fix"- we need to rewrite the connection layer a bit
             errMessage = "{\"message\": \"Connection Timeout\"}";
-          } else if (exception instanceof RxWebSocketCallback.Failure) {
```
luciofm, line 120: If it is not being used, just remove it... If it is something temporary, add a comment indicating that.

app/src/main/java/chat/rocket/android/activity/MainPresenter.java
leonardoaramaki: Done!

app/src/main/java/chat/rocket/android/activity/MainPresenter.java
leonardoaramaki: I agree. I don't have intention to keep it this way. I agree it's bad and the actual codebase is ignoring exceptions wide spread. Just kept this way for the sake of rushing to release this PR for testing. The code needs much refining

app/src/main/java/chat/rocket/android/service/RocketChatWebSocketThread.java
leonardoaramaki: I'll change those :) Just kept this way to keep the same naming convention already being used

app/src/main/java/chat/rocket/android/service/observer/MethodCallObserver.java
leonardoaramaki: Ok then, I thought over that one and removed the commented code. :+1:

philipbrito (commented):
Why calling `FLog.setMinimumLoggingLevel(...)` [two times](https://github.com/RocketChat/Rocket.Chat.Android/blob/40a217f0e57bb8b094c328fe3b56bf45d94a0f7e/rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/RocketChatWidgets.java)? And why setting `FLog.ERROR` instead of `FLog.VERBOSE` like recommended by [the documentation](http://frescolib.org/docs/troubleshooting.html)? Any special case? Anyway, take a look on [Fresco/ShowcaseApplication](https://github.com/facebook/fresco/blob/master/samples/showcase/src/main/java/com/facebook/fresco/samples/showcase/ShowcaseApplication.java) for reference.

leonardoaramaki:
@filipedelimabrito I didn't notice it was duplicated but I can rollback this one. I set to Flog.ERROR because Flog.VERBOSE spams a hell lot the logs with mostly useless information.

philipbrito:
The `Flog.VERBOSE` is especially useful when looking which `controllers` are been used with the `drawee`. This is a special cases to analyze, for example, how users avatars are shown in a RecyclerView, allowing a better/deep analyses.

leonardoaramaki:
Oh, I see. I'm sorry @filipedelimabrito. I never used Fresco before, didn't know it :)
I'll rollback on my next push!

luciofm (approved):
Approving it..

Guys, lets try to keep PRs smaller and self contained, this one had a lot of stuff that was not related do the issue, like notifications channel (should be a new PR), CI fixes (another PR), etc...

PR #365, opened 2017-07-26
Added a coding style

luciofm:


philipbrito (commented):
So happy we have code style for Kotlin and tests too!!!

luciofm:
Kotlin and Tests area are still very small, but we should add more rules as we go

rafaelks (approved):

PR #372, opened 2017-08-01
Show avatar with username initials.

### LINKED ISSUE #369: Default avatar (with user initials) isn't being presented
Your Rocket.Chat.Android version: 1.0.16 (28)
Mobile device model and OS version: Sony Xperia E5 (F3311) - Android 6.0

**Steps to reproduce:**
- User opens a channel that some user doesn't have avatar image

**Expected results:**
- Show the avatar with initials, just like on the web or iOS (screenshot below)

**Screenshot of the bug**
screenshot_20170728-090711

**Screenshot of the expected results**
screen shot 2017-07-28 at 09 08 59

philipbrito:
@RocketChat/android

Closes #369


app/src/main/java/chat/rocket/android/helper/RocketChatUserAvatar.kt
```diff
@@ -1,17 +1,91 @@
...
+     * @param context The context.
+     * @return A drawable with username initials.
+     * @see getUsernameInitials
+     */
+    fun getTextDrawable(username: String, context: Context): Drawable {
+        val round = (4 * context.resources.displayMetrics.density).toInt()
+
+        return TextDrawable.builder()
+                .beginConfig()
+                .useFont(Typeface.SANS_SERIF)
```
rafaelks: Will it use the default one from the OS?

app/src/main/java/chat/rocket/android/helper/RocketChatUserAvatar.kt
```diff
@@ -1,17 +1,91 @@
...
+     */
+    fun getUserAvatarBackgroundColor(username: String): Int {
+        return COLORS[username.length % COLORS.size]
+    }
+
+    private val COLORS = intArrayOf(
+            0xFFF44336.toInt(), 0xFFE91E63.toInt(), 0xFF9C27B0.toInt(), 0xFF673AB7.toInt(), 0xFF3F51B5.toInt(),
+            0xFF2196F3.toInt(), 0xFF03A9F4.toInt(), 0xFF00BCD4.toInt(), 0xFF009688.toInt(), 0xFF4CAF50.toInt(),
+            0xFF8BC34A.toInt(), 0xFFCDDC39.toInt(), 0xFFFFC107.toInt(), 0xFFFF9800.toInt(), 0xFFFF5722.toInt(),
+            0xFF795548.toInt(), 0xFF9E9E9E.toInt(), 0xFF607D8B.toInt())
```
rafaelks: What do you think about creating unit tests for this class? You could test a few example names with a few different usernames. On iOS I tested a bunch of different cases: https://github.com/RocketChat/Rocket.Chat.iOS/blob/develop/Rocket.ChatTests/Views/AvatarViewSpec.swift.

app/src/main/java/chat/rocket/android/helper/RocketChatUserAvatar.kt
philipbrito: It will be rendered as expected on Application, since the `Typeface` was [added](https://developer.android.com/reference/android/graphics/Typeface.html) by the Android platform since API level 1.

app/src/main/java/chat/rocket/android/helper/RocketChatUserAvatar.kt
philipbrito: Of course! I'll do unit tests! Thank you guy for putting those thoughts.

app/src/main/java/chat/rocket/android/helper/RocketChatUserAvatar.kt
rafaelks: 👍

app/src/main/java/chat/rocket/android/helper/RocketChatUserAvatar.kt
rafaelks: 👍

app/src/main/java/chat/rocket/android/helper/OkHttpHelper.java
```diff
@@ -48,4 +63,28 @@ public static OkHttpClient getClientForWebSocket() {
...
+      public void onResponse(@NonNull Call call, @NonNull final Response response)
+          throws IOException {
+        if (!response.isSuccessful()) {
+          throw new IOException("Unexpected code: " + response);
+        } else {
+          contentType = response.header("Content-Type");
+        }
+      }
+    });
+    return contentType;
```
luciofm: This most likely will be null, since you are doing an Asynchronous request.

app/src/main/java/chat/rocket/android/helper/RocketChatUserAvatar.kt
```diff
@@ -1,17 +1,91 @@
...
+
+    /**
+     * Returns the user avatar image format.
+     *
+     * @param uri The user avatar image URI to get the image format.
+     * @return The user avatar image format. Possible values are: "image/jpeg", "image/png" and "image/svg+xml".
+     * @see getUri
+     */
+    fun getImageFormat(uri: String): String {
+        return OkHttpHelper.getContentType(uri)
```
luciofm: We should have this information from the backend, each time we show an Avatar doing a HEAD (network operation) operation and a GET seems very expensive.

app/src/main/java/chat/rocket/android/helper/OkHttpHelper.java
```diff
@@ -1,21 +1,36 @@
...
 
 /**
  * Helper class for OkHttp client.
  */
 public class OkHttpHelper {
+  private static OkHttpClient defaultHttpClient;
   private static OkHttpClient httpClientForUploadFile;
   private static OkHttpClient httpClientForDownloadFile;
   private static OkHttpClient httpClientForWS;
+  private static String contentType;
```
luciofm: This will not work as expected...

First time you call `getContentType()` it will probably return `null`, since it is a async request.
Next time you call it, you will probably get the contentType from the previous request, which may be right or not.

On a list flinging, there will be dozens of requests overwriting `contentType`

app/src/main/java/chat/rocket/android/helper/RocketChatUserAvatar.kt
```diff
@@ -1,17 +1,91 @@
...
+                .buildRoundRect(getUsernameInitials(username), getUserAvatarBackgroundColor(username), round)
+    }
+
+    /**
+     * Returns a string with the username initials. For example: username John Doe returns JD initials.
+     *
+     * @param username The username.
+     * @return A string with username initials.
+     */
+    fun getUsernameInitials(username: String): String {
```
luciofm: The function doesn't conform with the documentation, it handles only separation by "." and not by spaces

app/src/main/java/chat/rocket/android/layouthelper/chatroom/AbstractMessageViewHolder.java
```diff
@@ -14,6 +14,7 @@
 public abstract class AbstractMessageViewHolder extends ModelViewHolder<PairedMessage> {
   protected final RocketChatAvatar avatar;
   protected final ImageView userNotFoundAvatarImageView;
+  protected final ImageView imageViewSvgUserAvatar;
```
luciofm: Do we need to use a separate ImageView? This should probably be handled on RocketChatAvatar.

app/src/main/java/chat/rocket/android/renderer/MessageRenderer.kt
```diff
@@ -19,18 +20,27 @@ class MessageRenderer(val message: Message, val autoLoadImage: Boolean) {
...
             rocketChatAvatarWidget.loadImage(message.avatar)
         } else {
             val username: String? = message.user?.username
             if (username != null) {
-                // Load user's avatar image from Rocket.Chat URI.
-                rocketChatAvatarWidget.loadImage(RocketChatUserAvatar(hostname, username).imageUri)
                 userNotFoundAvatarImageView.visibility = View.GONE
-                rocketChatAvatarWidget.visibility = View.VISIBLE
+                val userAvatarUri = RocketChatUserAvatar.getUri(hostname, username)
+                val userAvatarImageContentType = OkHttpHelper.getContentType(userAvatarUri)
```
luciofm: As commented before, this will not work consistently.

luciofm (changes requested):

app/src/main/java/chat/rocket/android/helper/OkHttpHelper.java
philipbrito: To get the `contentType` we need to do a network call - blocking operation, so we need to do an async task, right? There are the following possibilities: 

1) Get the `contentType` using the `HttpURLConnection ` with the `AsyncTask`.
2) Get the `contentType` using the `HttpURLConnection ` with `RxJava`.
3) Get the `contentType` using the `OkHttp ` which is currently used on RC app - `OkHttp` is a built-in response cache, a better cookie model, a better headers model and a better call model.

So, how we should handle this requirement?

app/src/main/java/chat/rocket/android/helper/RocketChatUserAvatar.kt
philipbrito: Yes, I pay attention w/ it, so to get the image format the requested method is `HEAD` and not `GET`. Take a [look](https://github.com/filipedelimabrito/Rocket.Chat.Android/blob/281ac8a39afa664b04cc3bf0e0674818f049682c/app/src/main/java/chat/rocket/android/helper/OkHttpHelper.java).

https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html:
> The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response. The metainformation contained in the HTTP headers in response to a HEAD request SHOULD be identical to the information sent in response to a GET request. This method can be used for obtaining metainformation about the entity implied by the request without transferring the entity-body itself. This method is often used for testing hypertext links for validity, accessibility, and recent modification.

rafaelks:
@filipedelimabrito Why don't you implement it the way iOS does?

Here's the structure of the AvatarView, basically:

```
-> UIView (the avatar view itself, root)
     -> ViewInitials (view with the initials, manually created, always present)
     -> ImageView (imageView with avatar, after loaded and cached)
```

This way, while the image is loading the imageView is transparent and the ViewInitials is the one that's being present to the user. If ImageView is loaded/cached, it will show the ImageView instead of ViewInitials.

app/src/main/java/chat/rocket/android/layouthelper/chatroom/AbstractMessageViewHolder.java
philipbrito: We need, because `RocketChatAvatar` is a class which returns a [SimpleDraweeView](http://frescolib.org/docs/using-simpledraweeview.html). There is a `setImageDrawable()` method on SimpleDraweeView but it is deprecated (trying to use it you'll get the messed avatars). Also, from [Fresco](http://frescolib.org/docs/gotchas.html):

> **Don’t set images directly on a DraweeView**
Currently DraweeView is a subclass of Android’s ImageView. This has various methods to set an image (such as setImageBitmap, setImageDrawable)
If you set an image directly, you will completely lose your DraweeHierarchy, and will not get any results from the image pipeline.

philipbrito:
That's I said before to you @rafaelks. We can set a placeholder image w/ username initials and if the image from backend is loaded successful then we show it. Note that the images w/ username initials are **loaded** from the backend, but it **isn't build** correctly - until now (see https://github.com/RocketChat/Rocket.Chat/pull/7572). As the chatroom background is white we can't see the username initials which are white too (the background isn't load because the SVG is build in a wrong way).

We are supporting the **SVG** rendering on RC app (I added it on some PR times ago). We simply can disable the SVG support to set the avatar w/ username initials and Fresco will do the rest (loading images that are not in SVG format, and setting the placeholder w/ username initials when the images are in SVG format - or, when the loading from server fails).

See this [comment](https://github.com/RocketChat/Rocket.Chat.Android/pull/372#discussion_r130892541) also, can be useful.

app/src/main/java/chat/rocket/android/helper/OkHttpHelper.java
philipbrito: > First time you call getContentType() it will probably return null, since it is a async request.

See the [comment](https://github.com/RocketChat/Rocket.Chat.Android/pull/372#discussion_r130874351) below.

> On a list flinging, there will be dozens of requests overwriting contentType

I agree w/ you. There is a way to avoid it, but it involves involves a disabled support. See [this comment](https://github.com/RocketChat/Rocket.Chat.Android/pull/372#issuecomment-319697445).

app/src/main/java/chat/rocket/android/helper/OkHttpHelper.java
luciofm: We should not do this, if we really need to get the Content-Type, we will need to implement our own Image loading library, like Fresco or Picasso, to download, cache and display it...

Doing a lot of `HEAD`s here will make a lot of useless requests on fling, we are not caching the result...

Ideally this should be handled on the Image library, on Fresco, or switch to some that already supports SVG.

app/src/main/java/chat/rocket/android/layouthelper/chatroom/AbstractMessageViewHolder.java
luciofm: My suggestion is to work with Fresco, and implement support for SVG directly on it...

luciofm: I mean, adding SVG support for Fresco

app/src/main/java/chat/rocket/android/helper/RocketChatUserAvatar.kt
filipedelimabrito: @luciofm Well... then I think we have a outdate documentation, because the [IOS version](https://github.com/RocketChat/Rocket.Chat.iOS/blob/develop/Rocket.Chat/Views/Avatar/AvatarView.swift) and [server version](https://github.com/filipedelimabrito/Rocket.Chat/blob/eefbd71089eac34e81d4f75f4e6c0dc07e1bd1a1/server/startup/avatar.js) only handles the separation by "." too.

Note that  we are getting the username and not the user's real name. It is right @rafaelks ?

rafaelks: @filipedelimabrito He's referring to the docs of your own method, I think.

rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/RocketChatWidgets.java
```diff
@@ -26,7 +31,7 @@ public static void initialize(Context context, OkHttpClient okHttpClient) {
     ImagePipelineConfig imagePipelineConfig = OkHttpImagePipelineConfigFactory
         .newBuilder(context, okHttpClient)
         .setRequestListeners(listeners)
-        .setImageDecoderConfig(CustomImageFormatConfigurator.createImageDecoderConfig())
+//        .setImageDecoderConfig(CustomImageFormatConfigurator.createImageDecoderConfig())
```
rafaelks: Can we remove this commented code?

rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/message/autocomplete/user/UserViewHolder.java
```diff
@@ -16,6 +20,12 @@
   private final RocketChatAvatar avatar;
   private final ImageView status;
 
+  private static final int[] COLORS = new int[] {
```
rafaelks: Why are you repeating the colors here?

rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/message/autocomplete/user/UserViewHolder.java
```diff
@@ -73,4 +83,35 @@ private String getImageUrl(String username, AbsoluteUrl absoluteUrl) {
...
+    int round = (int) (4 * context.getResources().getDisplayMetrics().density);
+
+    return TextDrawable.builder()
+        .beginConfig()
+        .useFont(Typeface.SANS_SERIF)
+        .endConfig()
+        .buildRoundRect(getUsernameInitials(username), getUserAvatarBackgroundColor(username), round);
+  }
+
+  private String getUsernameInitials(String username) {
```
rafaelks: This method seems to be duplicated...?

philipbrito: Yes :(
I just put it as before (w/ some the correct modifications). This class belongs to another module, there is the possibility to add the `:app` dependency to this module to avoid that duplication, but I really don't know if this is the correct (in the terms of design, architecture...). So I just did like before.

rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/message/autocomplete/user/UserViewHolder.java
philipbrito: Yes, for the same reason [bellow](https://github.com/RocketChat/Rocket.Chat.Android/pull/372#discussion_r131030142).

:(

rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/message/autocomplete/user/UserViewHolder.java
luciofm: I actually think the widgets should live on the app module... unless they are meant to be reusable, them they should live on another repository...

rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/RocketChatWidgets.java
philipbrito: If we need to plan to support SVG image in the future I think that we could leave this commented code. If not, we can remove it (and the CustomImageFormatConfigurator and SvgDecoder class too).

luciofm: We have github... We can always look the code history to recover this...

app/src/main/java/chat/rocket/android/helper/RocketChatUserAvatar.kt
philipbrito: @luciofm What did you mean?

rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/helper/UserAvatarHelper.kt
```diff
@@ -0,0 +1,98 @@
...
+
+        return TextDrawable.builder()
+                .beginConfig()
+                .useFont(Typeface.SANS_SERIF)
+                .endConfig()
+                .buildRoundRect(getUsernameInitials(username), getUserAvatarBackgroundColor(username), round)
+    }
+
+    /**
+     * Returns a string with the username initials. For example: username John Doe returns JD initials.
```
rafaelks: I think you should change this docs to something that will work. In your method `John Doe` will not become `JD`.

rafaelks (changes requested):
Just small change in your method docs.

app/src/main/java/chat/rocket/android/helper/RocketChatUserAvatar.kt
luciofm: Yes, the javadoc (for kotlin :smile:) says one thing and the function does another..

rocket-chat-android-widgets/src/test/kotlin/UserAvatarHelperTest.kt
```diff
@@ -0,0 +1,22 @@
+import chat.rocket.android.widget.helper.UserAvatarHelper
```
luciofm: Tests should live in the same package (and directory structure) as the class you are testing...

rocket-chat-android-widgets/src/main/res/layout/message_avatar.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <com.facebook.drawee.view.SimpleDraweeView xmlns:android="http://schemas.android.com/apk/res/android"
-        xmlns:fresco="http://schemas.android.com/apk/res-auto"
-        android:id="@+id/drawee_avatar"
-        android:layout_width="match_parent"
-        android:layout_height="match_parent"
-        fresco:actualImageScaleType="fitCenter"
-        fresco:roundedCornerRadius="5dp" />
\ No newline at end of file
+  xmlns:fresco="http://schemas.android.com/apk/res-auto"
```
luciofm: Format it with 4 spaces indentation, please.

rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/helper/UserAvatarHelper.kt
```diff
@@ -58,7 +58,7 @@ object UserAvatarHelper {
     }
 
     /**
-     * Returns a string with the username initials. For example: username John Doe returns JD initials.
+     * Returns a string with the username initials. For example: username John.Doe returns JD initials.
```
rafaelks: 👍

rafaelks:
LGTM, what about you @luciofm?

rafaelks (approved):

app/src/main/java/chat/rocket/android/helper/OkHttpHelper.java
```diff
@@ -48,4 +63,28 @@ public static OkHttpClient getClientForWebSocket() {
     }
     return httpClientForWS;
   }
+
+  public static String getContentType(String uri) {
```
luciofm: Can you remove this method that is not used anymore.

luciofm (approved):

PR #373, opened 2017-08-01
Update Google Play library version.

### LINKED ISSUE #360: Investigate notifications reliability on Android
Notifications seems to not be very reliable on Android. We also are using GCM which is deprecated now.

philipbrito:
@RocketChat/android

This PR updates the Google Play library version. 
Closes #360 ??

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -77,7 +77,7 @@ play {
 }
 
 ext {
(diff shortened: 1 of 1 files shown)


luciofm (approved):
:+1:

Still think we should move to FCM, just not high priority.

PR #377, opened 2017-08-04
Remove user not found handling.

### LINKED ISSUE #368: The list of messages show "User not found" before loading the user's username and avatar
Your Rocket.Chat.Android version: 1.0.16 (28)
Mobile device model and OS version: Sony Xperia E5 (F3311) - Android 6.0

**Steps to reproduce 01:**
- The app is closed
- User opens the app, waits for authentication
- On the same channel, the list of users will refresh to "User not found" and then a few seconds later change again to the correct username and avatar

**Steps to reproduce 02:**
- The app is opened
- User is authenticated
- User is looking a channel
- User changes to another channel
- The list of users will refresh to "User not found" and then a few seconds later change again to the correct username and avatar

**Screenshot**
screenshot_20170728-090208

philipbrito:
@RocketChat/android

Closes #368


luciofm (approved):
:+1:

PR #381, opened 2017-08-09
 Spotlight: search for users.

philipbrito:
@RocketChat/android

This PR adds the search for users on Spotlight. Currently, the Spotlight shows only rooms.

## Progress

- [x] Fetch users with the searched term.
- [x] Show the correct/related (at) icon for users.
 For example: 
      @ john.doe
      @ john.lennon
- [x] Create a DM with the selected user.
- [x] Unit tests.


luciofm:
Nice, will review it tomorrow morning.

philipbrito:
Thank you @luciofm !
It is in progress yet. I'll add a task list to show you the PR progress. Of course, feel free to review the files (code style, better code logic, etc).

geekgonecrazy:
With this change will it also search rooms you are already part of?

app/src/main/java/chat/rocket/android/fragment/sidebar/SidebarMainFragment.java
```diff
@@ -87,7 +83,7 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
     hostname = args == null ? null : args.getString(HOSTNAME);
 
     methodCallHelper = new MethodCallHelper(getContext(), hostname);
-    realmSpotlightRoomRepository = new RealmSpotlightRoomRepository(hostname);
+    realmSpotlightRepository = new RealmSpotlightRepository(hostname);
```
luciofm: We should move this to a presenter.

app/src/main/java/chat/rocket/android/api/MethodCallHelper.java
```diff
@@ -462,15 +466,43 @@ public MethodCallHelper(RealmHelper realmHelper, DDPClientRef ddpClientRef) {
...
   }
 
   public Task<Void> searchSpotlightRooms(String term) {
-    return searchSpotlight(
-        RealmSpotlightRoom.class, "rooms", term
-    );
+    return searchSpotlight(RealmSpotlightRoom.class, "rooms", term);
+  }
+
+  public Task<Void> searchSpotlight(String term) {
```
luciofm, line 476: I'd advocate to try to not spread more `Task`'s on the code base... Since this is a `Void` task, it could return a `Completable` and then it would be easy to compose a stream on the outer layers (Presenters, UI, etc).

On the future, I'd like to get rid of bolts tasks in favor of RxJava, so if we don't expose then, it will be easier to accomplish this.

philipbrito:
@geekgonecrazy Maybe not in this PR. Probably I'll create another PR w/ that improvement.

philipbrito:
Only refactoring before the merge...

geekgonecrazy:
> @geekgonecrazy Maybe not in this PR. Probably I'll create another PR w/ that improvement.

Ah ok.  I was getting excited.  Everyone I need to DM I've already done so in the past.  So not as exciting for me :grin:

But good to have the other stuff working

luciofm (approved):

philipbrito:
Merging after review and tests...

PR #388, opened 2017-08-16
Avoid Realm queries for messages with null users

leonardoaramaki:
@RocketChat/android

This PR fixes the flood of Realm queries for new messages. The queries' results are sent downstream only if changes have been observed and also it looks for messages that have actually a related RealmUser.
There's also a fix to avoid deleting some tables without notice (including the RealmUser one) upon reconnection. 
More changes includes:

- Implemented a ping call method that relies solely on RxJava. It was added not for the sake of transitioning from bolts but because it lacked reliability from the bolts' callbacks.
- Added a heartbeat method called at the end of  a (re)connection that is responsible for calling this new ping method and scheduling a keep-alive task for best ensuring the socket thread is kept alive after the process getting killed which was one source of socket drop without coming back issue.


android-ddp/src/main/java/chat/rocket/android_ddp/DDPClientImpl.java
```diff
@@ -106,40 +107,75 @@ public void connect(final TaskCompletionSource<DDPClientCallback.Connect> task,
...
-                          task.setResult(new DDPClientCallback.Ping(client, id));
-                          disposables.clear();
-                        }
-                      }
-                      disposables.clear();
+              .filter(response -> "pong".equalsIgnoreCase(extractMsg(response)))
+              .doOnError(error -> {
+                RCLog.e(error, "Heartbeat ping[%s] xxx failed xxx", id);
+              })
+              .map(response -> {
```
luciofm, line 124: I'm ok for now, but when you have complex mappings or `onNext` or `onError` handlings, extract to a meaningful method and use method reference like `this::processPingResponse`

android-ddp/src/main/java/chat/rocket/android_ddp/DDPClientImpl.java
```diff
@@ -106,40 +107,75 @@ public void connect(final TaskCompletionSource<DDPClientCallback.Connect> task,
...
+    if (requested) {
+      CompositeDisposable disposables = new CompositeDisposable();
+
+      disposables.add(
+              flowable.filter(callback -> callback instanceof RxWebSocketCallback.Message)
+                      .timeout(8, TimeUnit.SECONDS)
+                      .map(callback -> ((RxWebSocketCallback.Message) callback).responseBodyString)
+                      .map(DDPClientImpl::toJson)
+                      .subscribe(
+                              response -> {
```
luciofm, line 163: Same as above, for now it is ok, but for our new project, lets try to keep the code more readable, extracting complex `onNext` to meaningful methods.

android-ddp/src/main/java/chat/rocket/android_ddp/rx/RxWebSocket.java
```diff
@@ -62,14 +62,20 @@ public void onClosed(WebSocket webSocket, int code, String reason) {
               }
             }),
         BackpressureStrategy.BUFFER
-    ).delay(4, TimeUnit.SECONDS).publish();
+    ).publish();
```
luciofm, line 65: :beers: :clap:

app/build.gradle
```diff
@@ -134,6 +134,11 @@ dependencies {
 
     compile "com.github.hotchemi:permissionsdispatcher:$permissionsdispatcherVersion"
     annotationProcessor "com.github.hotchemi:permissionsdispatcher-processor:$permissionsdispatcherVersion"
+
+    provided 'com.parse.bolts:bolts-tasks:1.4.0'
```
luciofm: :+1: :clap:

app/src/main/java/chat/rocket/android/RocketChatCache.java
```diff
@@ -51,11 +53,11 @@ public String getOrCreatePushId() {
     return preferences.getString(KEY_PUSH_ID, null);
   }
 
-  public Flowable<String> getSelectedServerHostnamePublisher() {
+  public Flowable<Optional<String>> getSelectedServerHostnamePublisher() {
```
luciofm, line 56: Perfect, for the next project lets use more `Optional` or even more Kotlin

app/src/main/java/chat/rocket/android/activity/MainPresenter.java
```diff
@@ -98,8 +98,10 @@ public void onOpenRoom(String hostname, String roomId) {
 
   @Override
   public void onRetryLogin() {
-    view.showConnecting();
-    connectivityManagerApi.keepAliveServer();
+    final Disposable subscription = sessionInteractor.retryLogin()
+            .subscribe();
```
luciofm, line 102: Would be nice to have a logger `BiConsumer`, ignoring the result and logging errors.

app/src/main/java/chat/rocket/android/service/RealmBasedConnectivityManager.java
```diff
@@ -163,8 +165,8 @@ public void notifyConnecting(String hostname) {
       }
 
       return connectToServer(hostname)
-          //.doOnError(RCLog::e)
-          .retryWhen(RxHelper.exponentialBackoff(3, 500, TimeUnit.MILLISECONDS));
+          .doOnError(RCLog::e)
+          .retryWhen(RxHelper.exponentialBackoff(Integer.MAX_VALUE, 500, TimeUnit.MILLISECONDS));
```
luciofm, line 169: Doesn't `2147483647` a little too much?

luciofm (approved):

app/src/main/java/chat/rocket/android/service/RealmBasedConnectivityManager.java
leonardoaramaki: I agree it is. It's opened for discussion this max retry threshold. What I want to do on this one is set a high (not that high) one here and if we reach that max we defined then just display a RETRY button.

leonardoaramaki: Also I think on putting a counter there too, just like the web client, specially because we're using an exponential backoff retry time there, so it might take a while to try reconnecting again as time passes.

app/src/main/java/chat/rocket/android/service/RocketChatWebSocketThread.java
```diff
@@ -34,11 +36,15 @@
 import chat.rocket.android.service.observer.PushSettingsObserver;
 import chat.rocket.android.service.observer.SessionObserver;
 import chat.rocket.android.service.observer.TokenLoginObserver;
```
luciofm: This import is not used and are making pmd fail the build...

persistence-realm/src/main/java/chat/rocket/persistence/realm/repositories/RealmMessageRepository.java
```diff
@@ -135,13 +135,15 @@ public RealmMessageRepository(String hostname) {
         () -> new Pair<>(RealmStore.getRealm(hostname), Looper.myLooper()),
         pair -> RxJavaInterop.toV2Flowable(pair.first.where(RealmMessage.class)
             .equalTo(RealmMessage.ROOM_ID, room.getRoomId())
+            .isNotNull(RealmMessage.USER)
```
philipbrito, line 138: ;)

philipbrito (approved):

rafaelks:
Let's merge this one for now.

PR #389, opened 2017-08-16
Initial Crashlytics support

luciofm:


rafaelks (approved):

app/build.gradle
```diff
@@ -22,6 +29,7 @@ buildscript {
         classpath 'com.jakewharton.hugo:hugo-plugin:1.2.1'
         classpath 'com.google.gms:google-services:3.0.0'
         classpath 'com.github.triplet.gradle:play-publisher:1.1.5'
+        classpath 'io.fabric.tools:gradle:1.+'
```
philipbrito, line 32: I'm advocate to not using + in version numbers. Because... we know: `"[...] can lead to unpredictable and unrepeatable builds"`.

But it is how Fabric [documentation](https://docs.fabric.io/android/crashlytics/build-tools.html#set-up-a-library-subproject) says.

So, that's ok for me.

philipbrito (approved):

PR #392, opened 2017-08-16
Fix crash when building default avatar 

### LINKED ISSUE #390: Building default avatar with unexpected usernames with dots crashes the app
Your Rocket.Chat.Android version: 1.0.16

Mobile device model and OS version: All versions

Some unexpected usernames with dots causes crashes on building their avatars, such as "...", "..", ".Foo.", "Foo.bar."

leonardoaramaki:
@RocketChat/android

Closes #390 

Crashes were happening when building the default avatar for users with unexpected dotted usernames such as "...", "..", ".Foo." or "Foo..".

This changes the behavior to get close to the web client. The above mentioned usernames will be rendered, respectively, as "..", "..", ".F" and "FO". The web client does almost the same with few differences, like when the username is "..." it renders as an empty avatar.

### DIFF
--- a/rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/helper/UserAvatarHelper.kt
+++ b/rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/helper/UserAvatarHelper.kt
@@ -69,8 +69,11 @@ object UserAvatarHelper {
         }
 
         val splitUsername = username.split(".")
-        if (splitUsername.size > 1) {
-            return (splitUsername[0].substring(0, 1) + splitUsername[splitUsername.size - 1].substring(0, 1)).toUpperCase()
+        val splitCount = splitUsername.size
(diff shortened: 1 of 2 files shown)


rafaelks:
Looking good!!!

luciofm (approved):

philipbrito:
@RocketChat/android 

Guys, shall we expect at least all the requested revision to be made?
Too bad :( the jumping... since the review was requested one hour ago.

We are yet getting one small error, because we aren't comparing the `strings` properly on our test.

Take a look: https://github.com/RocketChat/Rocket.Chat.Android/pull/393

leonardoaramaki:
@filipedelimabrito Is it an issue on the tests you mean?

philipbrito:
Yep! But it was a small issue @leonardoaramaki 
I was almost to send a commit here, but I already [fixed](https://github.com/RocketChat/Rocket.Chat.Android/pull/393) it. :)

PR #393, opened 2017-08-16
Fix on building the default avatar

philipbrito:
@RocketChat/android

We are facing a small issue on building the default avatar. This PR fix it. It also updates the related test class (note that before this PR, the test are fine even testing `"Hi"` w/ `"hello"` string).

### DIFF
--- a/rocket-chat-android-widgets/src/test/kotlin/chat/rocket/android/widget/helper/UserAvatarHelperTest.kt
+++ b/rocket-chat-android-widgets/src/test/kotlin/chat/rocket/android/widget/helper/UserAvatarHelperTest.kt
@@ -1,26 +1,27 @@
 import chat.rocket.android.widget.helper.UserAvatarHelper
 import org.junit.Test
+import kotlin.test.assertEquals
 
 class UserAvatarHelperTest {
 
     @Test
     fun getUsernameInitialsTest() {
-        assert(UserAvatarHelper.getUsernameInitials("") == "?")
-        assert(UserAvatarHelper.getUsernameInitials("?") == "?")
-        assert(UserAvatarHelper.getUsernameInitials("f") == "F")
-        assert(UserAvatarHelper.getUsernameInitials("B") == "B")
-        assert(UserAvatarHelper.getUsernameInitials("fo") == "FO")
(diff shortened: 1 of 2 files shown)


rocket-chat-android-widgets/src/test/kotlin/chat/rocket/android/widget/helper/UserAvatarHelperTest.kt
```diff
@@ -1,26 +1,27 @@
...
+        assertEquals("B", UserAvatarHelper.getUsernameInitials("B"))
+        assertEquals("FO", UserAvatarHelper.getUsernameInitials("Fo"))
+        assertEquals("FO", UserAvatarHelper.getUsernameInitials("FO"))
+        assertEquals("FO", UserAvatarHelper.getUsernameInitials("fOo"))
+        assertEquals("FO", UserAvatarHelper.getUsernameInitials("FOO"))
+        assertEquals("FO", UserAvatarHelper.getUsernameInitials("F.O"))
+        assertEquals("FO", UserAvatarHelper.getUsernameInitials("F.o"))
+        assertEquals("FB", UserAvatarHelper.getUsernameInitials("Foo.bar"))
+        assertEquals("FB", UserAvatarHelper.getUsernameInitials("Foobar.bar"))
+        assertEquals("FZ", UserAvatarHelper.getUsernameInitials("Foobar.bar.zab"))
```
philipbrito, line 21: screen shot 2017-08-16 at 21 08 42

rafaelks: 👍

leonardoaramaki:
The test failed indeed and needs this fix. The behavior of getting the first and second initials from the first and second words between the dots was intentional though. Did this way because I thought I saw an edge case with this solution of taking always the second initial from the last word after the dot, but I can't recall now, maybe I was wrong :+1: 
It works for me this way :)

rafaelks (approved):

rafaelks:
Let's merge this one? @luciofm @leonardoaramaki

PR #408, opened 2017-08-23
[NEW] Show user status on toolbar

philipbrito:
@RocketChat/android

This PR shows the user status on toolbar (when you are in a DM). This reflects the status of the user whenever there is any change.


rafaelks:
Is this done @filipedelimabrito and @luciofm?

philipbrito:
I'm sorry. Since this PR shows the user status on toolbar, I took the advantage to improve (?) its UI.

[demo/temp image]
photo_2017-08-26_05-16-23

rafaelks:
@filipedelimabrito Found a few issues while testing your build...

1. If I'm a channel or a private group and I go to a offline user, the icon doesn't change.
2. Can you move just a little bit to the left the icon + name of the channel? The space is too big IMHO.
3. On my device the "bubble" seems too big... can we decrease the size by ~30%?

But it's looking great! Thank you! 👏

rafaelks (changes requested):
Just need to fix a few issues that I commented before.

philipbrito:
For sure @rafaelks!
1. -> Got it right now :/
2 and 3 -> Thanks for the feedback (always welcome). I'll fix it too.

Another UI changes will be on another PR (for better sanity).

rafaelks (approved):

PR #414, opened 2017-08-29
[FIX] Change menu icon color to white and the badge color

philipbrito:
This PR changes the navigation icon color (it was black and now it is white). Also it updates the badge color (from white to red).


luciofm (approved):
:+1:

PR #415, opened 2017-08-29
[FIX] Fix the screen dimming when sliding in rooms with one message

philipbrito:
I got a situation that the screen is not dimming totally. It happens when you see only one message on a chat room. Take a look:

photo_2017-08-28_23-03-30

Now:

photo_2017-08-28_23-03-34


rafaelks:
Good finding! Thanks @filipedelimabrito!

rafaelks (approved):

PR #430, opened 2017-08-31
[CHORE] Realm Stetho - Increase table view data limit

JigneshWorld:
@RocketChat/android As there was limited view of table while debugging realm database
https://github.com/uPhyca/stetho-realm/issues/2


luciofm:
Thanks for the PR @JigneshPatel23, will take a look on how it perform with such a big limit...

JigneshWorld:
Sure @luciofm. FYI I have checked with approx 1000 data in one of the table till then it works fine and it will be applicable while debugging only.

rafaelks:
@JigneshPatel23 Thanks for your PR! 👍

JigneshWorld:
@rafaelks Thanks.

PR #436, opened 2017-09-01
[NEW] Support multi-server

leonardoaramaki:
@RocketChat/android

Multi-server feature is available from a sliding panel to the left. I made use of the preexisting layout already mocked for this feature and replaced it. Needs more work on UI/UX for this to be more easy to spot and use.

Currently it tries to fetch the server logo from the server and put a placeholder just like the user's avatar to fallback from it. Unfortunately those logos are usually (if not always) an SVG file which the app fails to render. Since the fallback avatar is not very descriptive I put the hostname on the right side of the icon. Ideally it should be the site/server name, an information we can get from the Public Settings, but so far I've seen this setting is always equal to "Rocket Chat" on all tested servers so I though it was useless at this very moment.

This feature is pretty much a WIP since I think it's no stable enough and should be further tested but it's a great step forward.


rafaelks:
@leonardoaramaki This release is looking amazing, thanks a lot! 👍

geekgonecrazy:
This is great!  Just a note this doesn't seem to allow notifications for the other servers.  I've added 4 and only one gets notifications

rafaelks:
Everything is looking very good, I would only suggest two more changes to the feature before we merge it:

1) Use setting `Site_Name` to show the server name
2) Change the "pressing" area into the button to all width (image below)

screenshot_20170905-053942

rafaelks:
@leonardoaramaki Found two bugs when testing the "upgrade" of version (from Beta).

1) Delete the app;
2) Install from Google Play (Beta);
3) **Bug:** asked me directly username/password. The app is fresh new, should as URL first;
4) Authenticated in demo, OK;
5) Now install this build from CircleCI that has the latest changes in this PR (https://951-48179404-gh.circle-artifacts.com/0/tmp/circle-artifacts.H7UKp1z/app-release.apk)
6) **Bug:** opens the app, everything is blank, nothing shows;
7) Closes the app and opens again, it's all OK;
8) Servers are showing on the left, OK;

rafaelks (approved):

PR #448, opened 2017-09-08
[FIX] Set client to use Cordova's project Sender Id when no token on Settings

leonardoaramaki:
@RocketChat/android

Fallback to using Cordova's sender id when no other on is set under the admin settings for the server.


rafaelks (approved):

luciofm (approved):
:+1:

philipbrito (approved):

PR #449, opened 2017-09-08
[FIX] Spotlight results

philipbrito:
@RocketChat/android

This PR changes a little bit the Spotlight logic. If the inputed text doesn't match your subscribed rooms then there is no needed to tap over "Load more results" to show the rooms that you are not subscribed (to allow you the possibility to subscribe them). 

If text matches, you have the option to tap over "Load more results" to show unsubscribed rooms for the corresponding query term too.


luciofm (approved):
:+1:

leonardoaramaki (approved):

Issue #424, opened 2017-08-30
[NEW] Allow user to see the files list of a channel/group/dm


rafaelks:
User needs to be able to see the files list of some channel. There's a REST API (paginated) for that: https://github.com/RocketChat/Rocket.Chat/pull/7793/files.

PR #454, opened 2017-09-11
[NEW] Room's menu (REST API calls)

### LINKED ISSUE #356: [BUG] Opening member list from channels with very large amounts of members crashes
Your Rocket.Chat.Android version: 1.0.16

To reproduce this just open the member list from channel **general** which has more than 100k members.

This crash is caused by an OutOfMemoryError since during the loading the app tries to build a json object putting every member info inside it. 

A solution to this would be lazy loading this list.

### LINKED ISSUE #421: [NEW] Allow user to see the members list of a channel/group
User needs to be able to see the members list of some channel. There's a REST API (paginated) for that: https://github.com/RocketChat/Rocket.Chat/pull/7793/files.

### LINKED ISSUE #422: [NEW] Allow user to see the favorite messages list of a channel/group/dm
User needs to be able to see the favorite messages list of some channel. There's a REST API (paginated) for that: https://github.com/RocketChat/Rocket.Chat/pull/7793/files.

### LINKED ISSUE #423: [NEW] Allow user to see the pinned messages list of a channel/group/dm
User needs to be able to see the pinned messages list of some channel. There's a REST API (paginated) for that: https://github.com/RocketChat/Rocket.Chat/pull/7793/files.

philipbrito:
@RocketChat/android

Closes #421 
Closes #422 
Closes #423 
Closes #356 

It also serves as a start point to add other functionalities to the app (for example: leave a room).

All the features are implemented using the [REST API calls](https://rocket.chat/docs/developer-guides/rest-api/).

**UPDATE**: #424 will not close because it is necessary to make changes to the corresponding API, although the most important stuff to implement this functionality on the Android app already is done by this PR.

## Progress

- [x] Show member list of a room (and its total of members).
     - [x] For DM
     - [x] For Private Channels (aka groups)
     - [x] For Public Channels (aka channels)
- [x] Show favorite messages list of a room.
     - [x] For DM
     - [x] For Private Channels (aka groups)
     - [x] For Public Channels (aka channels)
- [x] Show pinned messages of a room.
     - [x] For DM
     - [x] For Private Channels (aka groups)
     - [x] For Public Channels (aka channels)
- [ ] Show files list of a room. **Need REST API fixes**
     - [ ] For DM
     - [ ] For Private Channels (aka groups)
     - [ ] For Public Channels (aka channels)
- [x] Implement a lazy loading list.
- [x] Unit tests.

photo_1 photo_1


app/src/main/java/chat/rocket/android/fragment/chatroom/RoomFragment.java
```diff
@@ -303,37 +299,40 @@ public boolean onItemLongClick(PairedMessage pairedMessage) {
...
     });
 
-    DrawerLayout drawerLayout = rootView.findViewById(R.id.drawer_layout);
-    if (drawerLayout != null && pane != null) {
-      compositeDisposable.add(RxDrawerLayout.drawerOpen(drawerLayout, GravityCompat.END)
-          .compose(bindToLifecycle())
-          .subscribe(
-              opened -> {
-                try {
-                  Field fieldSlidable = pane.getClass().getDeclaredField("mCanSlide");
```
philipbrito, line 323: Goodbye reflection!!!!!!!!!!

👋

rafaelks:
@filipedelimabrito Thanks! Looking good! 👍 

I did found a few UI issues that I believe we could improve (can be in another pull-request, no problem).

1. Members list cell row has too much space in the top; The space in the bottom & top should be the same IMHO;
2. The space between the member avatar & the status view (rounded colored) is too small, maybe something like 8pt would fit better; Here we could use the same horizontal space between status view & name/username label;
3. Favorited/Pinned messages doesn't show the image/attachment; I think we're not reusing the message cell here, I think we should;

Screenshots:

screenshot_20170921-090111
screenshot_20170921-090736
screen shot 2017-09-21 at 9 07 48 am

philipbrito:
(1) and (2): Fixed. I just changed the status view to another place on the screen. It is important to note that it follows the Material guidelines (as shown in the image below).

photo_2017-09-22_23-40-00

photo_2017-09-22_23-40-03

philipbrito:
(3): I added the widgets to show the images and attachments, but our REST API need a better design because there is a lot of loop to do to get a simple information, so it is bad.

We could improve that.

Example:

```
[...]
"urls":[
            {
               "url":"http:\/\/127.0.0.1:3000\/",
               "meta":{
                  "pageTitle":"Sign in | Patched Edition"
               },
               "headers":{
                  "contentType":"text\/html; charset=utf-8"
               },
               "parsedUrl":{
                  "host":"127.0.0.1:3000",
                  "hash":null,
                  "pathname":"\/users\/sign_in",
                  "protocol":"http:",
                  "port":"3000",
                  "query":null,
                  "search":null
               }
            },
[...]
```

philipbrito:
Also there is some missing information from the Rest: The user status, its name and offSet proprieties doesn't come from `/api/v1/channels.messages`, `/api/v1/groups.messages`  and `/api/v1/dm.messages`.

filipedelimabrito:
Close #356

rafaelks:
Looking great @filipedelimabrito!!! Thanks!

I can reproduce one crash only: if I go to the list of members and tap "Back" before the list loads, it crashes after a few seconds to me.

philipbrito:
Thanks for reporting this crash @rafaels!

I just add some commits that fixes that! Also, I did some improvements in order to a better performance when viewing a large list (i.e. with many itens to show).

rafaelks (approved):

philipbrito:
#424 closed from #491

PR #468, opened 2017-09-19
[FIX] Logout from single instance

leonardoaramaki:
@RocketChat/android

Logout from a single instance when multiple servers are added.


rafaelks:
@leonardoaramaki Found a bug while testing the app: if I'm in ~3 servers and I logout from one, it shows the me server screen after a while. Should only ask for a new server if I don't any server anymore.

rafaelks (approved):

PR #494, opened 2017-09-30
[FIX] Fix a deadlock on RocketChatService causing some ANR errors

luciofm:
- In the case of RocketChatWebSocketThread.getStarted() returning an Error, we never release the webSocketThreadLock.


philipbrito (approved):

leonardoaramaki (approved):
Nice catch!

PR #512, opened 2017-10-06
[NEW] Add Reply feature

leonardoaramaki:
@RocketChat/android

Adds the **reply** feature. All message actions, including **editing** are now available at a context menu which is popped up with a single tap on the message. 
A simple ui has been designed to signalize that a reply action is taking place by pulling up a banner/bar over the message composer. This banner shows a small thumb if the message being replied is an image. 
Take a note though that currently, to reply to an image message, the user has to tap outside the image boundaries, because doing otherwise the action of previewing the image will take place instead of opening the contextual menu. Later on, I plan to add an action menu triggered by long pressing the message containing those same actions which are available at the context menu.

**UPDATE**
The context menu containing all the actions on messages is now access by long pressing the desired message.


rafaelks:
Looking great!!! Small suggestions:

1. Only show contextual menu on long press, not on tap;
2. When you cancel a reply (by tapping in the X button), let's remove the focus from the keyboard;

app/src/main/java/chat/rocket/android/fragment/chatroom/RoomContract.java
```diff
@@ -58,5 +66,9 @@
     void onMarkAsRead();
 
     void refreshRoom();
+
+    void replyMessage(Message message);
+
+    void copyMessage(Message message);
```
philipbrito, line 72: Cool! That PR includes a copy message feature, right?
Thank you @leonardoaramaki!

philipbrito (approved):

app/src/main/java/chat/rocket/android/fragment/chatroom/RoomContract.java
leonardoaramaki: Yes @filipedelimabrito :) 
We just need to improve this when it comes to copy message replies! But it's already doing it's work :+1:

rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/message/MessageFormLayout.java
```diff
@@ -173,6 +208,61 @@ public void setEditTextCommitContentListener(
     this.listener = listener;
   }
 
+  public void setReplyContent(@NonNull AbsoluteUrl absoluteUrl, @NonNull Message message) {
+    String text = message.getMessage();
+    replyUsernameText.setText(message.getUser().getUsername());
```
geekgonecrazy, line 213: how hard would it be to add a check here and not add this if username is same as self?

PR #522, opened 2017-10-12
[NEW] Added support for Delete message #347

### LINKED ISSUE #347: [NEW] Delete message
Your Rocket.Chat.Android version: 1.0.15

The app should allow the message deletion. Currently only message edition is allowed to the user (holding down its message).

dimsedane:
@RocketChat/android

Closes #347 

Adds a delete action to the message context menu

screenshot_1507790552


dimsedane:
I have now tried multiple times to sign the CLA, but the assistant keeps saying I have not signed it. Any advise?

philipbrito:
@dimsedane The CLA is already signed! I'll review your PR soon! Thank you!

filipedelimabrito:
@dimsedane I tested your PR, but it isn't deleting the message. After tapping the delete option the message is hidden (not deleted) from the chat room (you can see after changing the room and coming back to the last room where you "delete" the message and checking it on the server too). Also, the option is showing to every message on the room (it is expected that only the original message creator delete its message).

philipbrito (changes requested):
@dimsedane I tested your PR, but it isn't deleting the message. After tapping the delete option the message is hidden (not deleted) from the chat room (you can see after changing the room and coming back to the last room where you "delete" the message and checking it on the server too). Also, the option is showing to every message on the room (it is expected that only the original message creator delete its message).

dimsedane:
I have made a small update to the code now. The delete observer was observing both posts marked for delete and those it was deleting. Tested on our own server, and the message was deleted.

I have switched to use the same permission for delete as for edit since I assume that if you can edit a message you can also delete it. Please let me know if that assumption is wrong :)

philipbrito:
Thank you so much for your PR @dimsedane! This is working very well! 
Yes! Good assumption! :)

philipbrito (approved):

Issue #363, opened 2017-07-25
[NEW] Honor setting to show first name / last name


geekgonecrazy:
Reference to issue created on main repo: https://github.com/RocketChat/Rocket.Chat/issues/7573

There is a setting to display first name and last name instead of username.  The Rocket.Chat+ doesn't yet have that feature

danieljhochman:
+1

rafaelks:
Information to accomplish this task:

# Description

> When `UI_Use_Real_Name == true`

- Every part of the app that uses Subscription model will display fname property;
- Every part of the app that uses User model will display name property;

> When `UI_Use_Real_Name == false`

- Every part of the app that uses Subscription model will display name property;
- Every part of the app that uses User model will display username property;

Implementation of the setting in iOS: https://github.com/RocketChat/Rocket.Chat.iOS/pull/565.

rafaelks:
Closing now, because it's already implemented in [2.0.0](https://github.com/RocketChat/Rocket.Chat.Android/milestone/6).

PR #524, opened 2017-10-13
[DESIGN][NEW] Chat room screen improvements + Shows the user real name

philipbrito:
@RocketChat/android

This PR aims to improve the look & feel of the chat room screen reducing the existing "visual pollution" and following the Android Material Guideline convection.

You'll notice that  it is more easy to replay, quote and edit a specific message (and to delete after the #522 is merged and to star/unstar/pin/unpin when these features are implemented).

This PR is able to show the user real name after #526 is merged.

Note: This PR remove the bluish navigation bar.

Preview:
photo2


philipbrito:
@RocketChat/android Guys, we are able to show the user "real name" now. As you can see on the bellow image, there are user with the "real name" while there are other users with only the username. We can show only the user "real name" (when available) instead of its username. What do you think?

Related: https://github.com/RocketChat/Rocket.Chat.Android/issues/363
photo_2017-10-14_13-51-20

rafaelks:
@filipedelimabrito There's a setting that tell wether the clients should show the username or user's real name. We should respect this setting.

philipbrito:
Where can I found that setting @rafaelks ?

rafaelks:
You can check how it works here: https://github.com/RocketChat/Rocket.Chat.Android/issues/363#issuecomment-336652280.

But let's leave this for **2.0.0** only.

rafaelks:
@filipedelimabrito Let's keep it closed for now, doesn't want go crazy with this change at this point, just adding what's necessary.

philipbrito:
@rafaelks @RocketChat/android 

Good job lost. This change only impacts the UI, I think that it doesn't crazy us (the users). Anyway, if someone want this "app version" go to my branch. Thanks

PR #526, opened 2017-10-13
[FIX] Missing name field in Realm database.

estevanfick:


philipbrito (approved):
Cool! Thanks for your contribution @estevanfick!

I just checked if the name is shown with #524 and it works!

philipbrito:
@RocketChat/android Because it changes the `/ddp/RealmUser.java` there will be a new login request.

PR #535, opened 2017-10-17
Closes #342. Adds Play Store update check and update button to sideba…

### LINKED ISSUE #342: [NEW] Suggest app update when new version is available
According to Google Play Console, 1/3 of our user base is still using older versions of the app.
As a counterpart, we may suggest users to upgrade whenever it's needed.

Libs like this one https://github.com/javiersantos/AppUpdater checks for the latest version available on GooglePlay (and another websites) and provide callbacks in order to communicate with the users. 

For UX, I'd suggest adding a label on the side menu, alerting if an upgrade is available. When clicked, the user gets redirected to the corresponding app store. Users should not be prevented from using the application, if they're not using latest.

JoshuaJamesOng:
…r with link to Play Store listing. Adds simple notification to toggle.

@RocketChat/android

Closes #342 

## Description
As per 342, uses AppUpdater library to check Play Store version name and adds menu item if update available. Also adds blue notification dot to indicate to user.  

## Implementation
`SidebarMainPresenter` subscribes to `CheckUpdateInteractor` `Observable`. The interactor upon subscription forces a re-fresh of the `UpdateRepository` and will return a `Boolean` as to whether the Play Store version is greater than the current (as per library logic). In the case of an error, it will assume there is no update. Based on the subscribed value, the view will either be invoked to show or hide the "Update RocketChat" menu item (hidden by default) and draw a `color_accent` circle to the `DownUpToggleView` `Canvas`. Implementation can also handle the repository checking at an interval - currently it will only check when a view is bound to the sidebar presenter. 

## Screenshots
screenshot_20171017-225049
screenshot_20171017-225055

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -167,6 +167,7 @@ dependencies {
     compile 'com.jakewharton.timber:timber:4.5.1'
     compile 'com.github.matrixxun:MaterialBadgeTextView:c5a27e8243'
     compile 'com.github.chrisbanes:PhotoView:2.0.0'
+    compile 'com.github.javiersantos:AppUpdater:2.6.3'
     provided 'io.reactivex:rxjava:1.3.0'
     provided "com.github.akarnokd:rxjava2-interop:0.10.2"
     provided 'com.hadisatrio:Optional:v1.0.1'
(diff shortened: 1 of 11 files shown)


philipbrito:
@JoshuaJamesOng Thanks for your contribution!
Can you fix your PR so that we could test it?

PR #547, opened 2017-11-01
[V2][LAYOUT] Authentication screens.

### LINKED ISSUE #544: [NEW][REWRITE] Authentication screens
# Description

 Write the layout for our authentication process.

## Progress

- [x] Connect to the server
- [x] Authentication
  - [x] Auth with username / email
  - [x] Signup with email
  - [x] Login/Signup with social networks
  - [x] 2FA code screen
  - [x] Register username

## Some things to pay attention

- The colors needs to be in a separated file, to be shared all over the app and easy to changed;
- All strings needs to be i18n;

philipbrito:
This PR adds the authentication screens for v2.
Closes #544 

Note:
- I took advantage and added the file that contains the translation for Brazilian Portuguese (default is english), so community, you can do the same and add another translation (like to French and German language :smile: ) .
- The activities and fragments are in the app directory, but it should be in a better place (let's do that after the project architecture definition).

Check out demo: https://youtu.be/ro9L-L3uHQk

TODO:

- [x] Up layout when keyboard is show. 
- [x] Design the Oauth/Social accounts interface.


dependencies.gradle
```diff
@@ -5,7 +5,7 @@ ext {
         java              : JavaVersion.VERSION_1_8,
         compileSdk        : 26,
```
leonardoaramaki: We should change compileSdk, targetSdk to 27 and buildTools as well as support to 27.0.0 so we can use the latest apis.

PR #565, opened 2017-11-14
[PROJECT][V2]Add Fresco library.

philipbrito:
@RocketChat/android

Also rearrange the dependencies order (by groupId).


luciofm:
Do we really need fresco??? We are not a image based app, and fresco have terrible APIs...

luciofm:
@leonardoaramaki what do you think?

philipbrito:
@luciofm Yes, we need! For example, to download the user avatar image... 
Fresco is considered the best image management library.
Fresco is getting more and more reliable (it comes from the Facebook company).
We are using Fresco in the `V1` and I personally don't have any problems with its API. 

Interesting reading: https://stackoverflow.com/a/29567946/4744263
https://medium.com/android-dev-br/fresco-sim-cda40fabae82 (PT-BR)

philipbrito:
@luciofm @leonardoaramaki Also, the Fresco team created an app that shows the "perfomance of five libraries - Fresco, Picasso, UIL, Glide, and Volley Image Loader - side by side" (Fresco member's words).

luciofm:
> Yes, we need! For example, to download the user avatar image...

But it can be done with other libraries, like glide or picasso or any other...

> it comes from the Facebook company

Is that a good thing?? :stuck_out_tongue: :scream: 

We are not an image based, or image heavy app, I'd prefer a better API over a small performance gain...

philipbrito:
> But it can be done with other libraries, like glide or picasso or any other...

If there is a library better than another (already proven) why not use it?

> Is that a good thing??

Yes, it is good (I got good things coming from [Lucas Rocha](https://github.com/lucasr) and others people that I had a chance to talk with). But I agree that it is merely an opinion (like Linux is better than Mac because...).

> We are not an image based, or image heavy app, I'd prefer a better API over a small performance gain...

Remember the PNG, SVG and GIF uploaded by the users in a chat room. Also, if we are not an image based app we will have no pain working with its API  :stuck_out_tongue: (again, I had no problem with Fresco API, sincerely).

To me, the key point is: If there is a library better than another (already proven) why not use it?
Just my thoughts.

leonardoaramaki:
@luciofm @filipedelimabrito Personally, I prefer using a simpler yet solid library like Picasso. I think that this MVC architecture of Fresco is like a cannon to smash an ant. :)

philipbrito:
c/c @luciofm 
@leonardoaramaki It is not true! The Fresco architecture is powerful. You can see through this line:

> "This modularization allows the base Fresco library to be light-weight. Adding those additional libraries would account for ~100 KiB each."

Source: http://frescolib.org/docs/faq.html

Also, are we planning to maintain the Fresco library too? How is the impact by adopting that library in the `v2`? It will break our architecture (reactive?). I can't see that, because we'll use a few lines of the Fresco, like:

```
Uri uri = Uri.parse(url);
draweeView.setImageURI(uri);
```

philipbrito:
c/c @leonardoaramaki 
@luciofm Because you are a Square fan, just another comparison between Fresco and Picasso: 

- Picasso:
Pom size: 2KB
Jar size (sources): 71 KB

- Fresco:
Pom size: 1KB
Jar size (sources):  11KB

Are we planning to maintain the Fresco (Picasso or whatever) library or to have an app to download, cache, display (etc.) an image as faster as possible?

PR #581, opened 2017-11-21
[FIX] servers logging out on app update

leonardoaramaki:
@RocketChat/android


rafaelks:
@leonardoaramaki I installed `1.0.22`, logged in a server and then install this APK on top of it, I was logged out again.

PR #621, opened 2017-12-06
[IMPROVEMENT] SVG support

### LINKED ISSUE #618: [NEW] Support SVG images
The app should support SVG images in order to show the avatar, server icon and user uploaded images with that format.
We should analyse the app performance when showing that images (to ensure that there is not high memory allocation and high CPU usage for example).

philipbrito:
This PR adds the SVG support for the app. 

Here is a document (just for reference) that shows the RC v2 app performance measurements proving that there are no major impacts when supporting this image format: https://goo.gl/y9aWPm 

Closes #618


leonardoaramaki:
@filipedelimabrito Does the app can now render this svg -> [https://thedevconf.rocket.chat/assets/logo.svg](url)?

philipbrito:
@leonardoaramaki Yes, the app must have the ability to open any SVG image. We need only to pay attention with the `header tag` of the SVG file. For instance, this [image](https://open.rocket.chat/avatar/hcharles) starts with `<?xml` header tag, while this [image](https://rocket.chat/images/index/features.svg) starts with `<svg` header tag. In both cases that supported `header tags` are declared [here](https://github.com/filipedelimabrito/Rocket.Chat.Android/blob/d176a33942d47d7b3d1ad6e6f380b2e93f9bd26f/app/src/main/java/chat/rocket/android/app/utils/SvgDecoder.kt#L26) and [here](https://github.com/filipedelimabrito/Rocket.Chat.Android/blob/d176a33942d47d7b3d1ad6e6f380b2e93f9bd26f/app/src/main/java/chat/rocket/android/app/utils/SvgDecoder.kt#L27).

leonardoaramaki:
@filipedelimabrito Nice. Have you/Can you test with both images? I ask this because we had something similar to that on legacy and still Fresco integration with [https://github.com/BigBadaboom/androidsvg](url) didn't work as expected

PR #622, opened 2017-12-07
[WIP] Initial SDK integration and Login support

luciofm:
Integrated the SDK and add login support (still missing save the credentials do the DB, verify if the user is logged in when starting the app, and cleanups)


app/src/main/java/chat/rocket/android/authentication/presentation/AuthenticationNavigator.kt
```diff
@@ -0,0 +1,34 @@
...
+    var currentServer: String? = null
+
+    fun toLogin(server: String) {
+        currentServer = server
+        activity.addFragmentBackStack("loginFragment", R.id.fragment_container) {
+            LoginFragment.newInstance(server)
+        }
+    }
+
+    fun toChatRoom() {
```
philipbrito: Prefer `toChatList()` instead of `toChatRoom` because the (user) chat list is the first screen shown to the user after authentication.

dependencies.gradle
```diff
@@ -30,13 +32,16 @@ ext {
...
+        kotlin              : "org.jetbrains.kotlin:kotlin-stdlib-jre8:${versions.kotlin}",
+        coroutines          : "org.jetbrains.kotlinx:kotlinx-coroutines-core:${versions.coroutine}",
+        coroutinesAndroid   : "org.jetbrains.kotlinx:kotlinx-coroutines-android:${versions.coroutine}",
 
         appCompat           : "com.android.support:appcompat-v7:${versions.support}",
         annotations         : "com.android.support:support-annotations:${versions.support}",
         recyclerview        : "com.android.support:recyclerview-v7:${versions.support}",
         design              : "com.android.support:design:${versions.support}",
         constraintLayout    : "com.android.support.constraint:constraint-layout:${versions.constraintLayout}",
+        cardView            : "com.android.support:cardview-v7:${versions.support}",
```
philipbrito, line 44: Are we using the cardView in some layout? If not, we don't need to declare this dependency for now.

app/build.gradle
```diff
@@ -14,30 +14,43 @@ android {
...
+
     implementation libraries.kotlin
+    implementation libraries.coroutines
+    implementation libraries.coroutinesAndroid
 
     implementation libraries.appCompat
     implementation libraries.recyclerview
     implementation libraries.design
     implementation libraries.constraintLayout
+    implementation libraries.cardView
```
philipbrito, line 44: Are we using the cardView in some layout? If not, we don't need to declare this dependency for now.

app/src/main/java/chat/rocket/android/util/ActivityExtensions.kt
```diff
@@ -0,0 +1,14 @@
+package chat.rocket.android.util
+
+import android.support.v4.app.Fragment
+import android.support.v7.app.AppCompatActivity
+
+fun AppCompatActivity.addFragment(tag: String, layoutId: Int, newInstance: () -> Fragment) {
```
philipbrito, line 6: Good! 👍 
Anyway I'd recommend this reading: https://www.philosophicalhacker.com/post/how-to-abuse-kotlin-extension-functions/

dependencies.gradle
luciofm: CardView is a dependency of FloatingSearchView, but they are using version 25.3.1, which causes lint errors.

luciofm: @filipedelimabrito btw, had you take a look on alternatives to this library? The last commit is 6 months older, I think we should look for a better maintained library

app/src/main/java/chat/rocket/android/authentication/presentation/AuthenticationNavigator.kt
luciofm: Makes sense...

dependencies.gradle
philipbrito: Yes, I took a look on alternatives. One of the alternatives is "SearchView" but unfortunately I still have no answer to an issue reported by me: https://github.com/lapism/SearchView/issues/339

Even not being actively maintained, the "FloatingSearchView" has more stargazers than other libraries and [explains](https://github.com/arimorty/floatingsearchview/wiki) in a commendable way how we can make change to our own copy from its code base.


We can decide later if we'll use our copy of the "FloatingSearchView" or we'll use the "SearchView" library, but it isn't highly relevant at the moment IMHO.

lapism: About my SearchView. I am still working on it and I will release next Beta version until Christmas.

philipbrito: Thanks @lapism for working in it! We hope that we can use SearchView when you release a new version addressing that issues!

Thanks for your time! :heart:

PR #623, opened 2017-12-07
[LAYOUT] Chat room interface

### LINKED ISSUE #615: [NEW] Screen that shows the messages list of a chat room
Initial implementation requires only list the text messages: avatar, user (username), message text & timestamp.

In the bottom of the screen, add a text box to let user write a message and tap "Send".

philipbrito:
This PR adds the chat room interface.

Closes #615


app/src/main/java/chat/rocket/android/app/DateTimeHelper.kt
```diff
@@ -35,6 +35,17 @@ object DateTimeHelper {
         }
     }
 
+    /**
+     * Returns a time from a LocalDateTime.
+     *
+     * @param localDateTime The LocalDateTime.
+     * @return The time from a LocalDateTime.
+     */
+    fun getTime(localDateTime: LocalDateTime): String {
```
luciofm, line 44: I'd vote to use the JSR-310 backport for time (https://github.com/JakeWharton/ThreeTenABP)

app/src/main/java/chat/rocket/android/app/User.kt
```diff
@@ -0,0 +1,7 @@
+package chat.rocket.android.app
+
+data class User(val id: String,
```
luciofm, line 3: We should move all this temporary model classes to a separate `model` package

app/src/main/java/chat/rocket/android/app/chatlist/ChatListAdapter.kt
```diff
@@ -22,20 +22,15 @@ class ChatListAdapter(private var dataSet: MutableList<Chat>, private val contex
     override fun onBindViewHolder(holder: ViewHolder, position: Int) {
         val chat = dataSet[position]
 
-        holder.userAvatar.setImageURI(chat.userAvatarUri)
+        holder.userAvatar.setImageURI(chat.user.avatarUri)
         holder.chatName.text = chat.name
         holder.lastMessage.text = chat.lastMessage
         holder.lastMessageDateTime.text = DateTimeHelper.getDate(chat.lastMessageDateTime, context)
 
         when (chat.type) {
```
luciofm, line 30: We should avoid any kind of processing on the Adapter... When we have the SDK in place, and the final models, we will convert from the DB model to a somekind of ViewModel that have all information needed for the adapter... The adapter work should be just binding information to views, not processing informations...

For now, with fake data this is ok, but lets keep that in mind

app/src/main/java/chat/rocket/android/app/chatroom/MessageFragment.kt
```diff
@@ -0,0 +1,38 @@
...
+        val user2 = User("2", "Lucio Maciel", "Lucio Maciel", "busy", "https://open.rocket.chat/avatar/lucio.maciel")
+
+        val message1 = Message(user1, "This is a multiline chat message from Bertie that will take more than just one line of text. I have sure that everything is amazing!", LocalDateTime.now())
+        val message2 = Message(user2, "Great!", LocalDateTime.now().plusHours(1))
+        return listOf(message1, message2)
+    }
+
+    // REMARK: The presenter should call this method.
+    private fun showMessageList(dataSet: List<Message>) {
+        recycler_view.layoutManager = LinearLayoutManager(activity.applicationContext, LinearLayoutManager.VERTICAL, false)
```
luciofm, line 35: Use the activity context

app/src/main/java/chat/rocket/android/app/chatroom/MessageListAdapter.kt
```diff
@@ -0,0 +1,37 @@
...
+import android.view.ViewGroup
+import android.widget.TextView
+import chat.rocket.android.R
+import com.facebook.drawee.view.SimpleDraweeView
+import kotlinx.android.synthetic.main.item_message.view.*
+
+class MessageListAdapter(private var dataSet: MutableList<Message>) : RecyclerView.Adapter<MessageListAdapter.ViewHolder>() {
+
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): ViewHolder {
+        val view = LayoutInflater.from(parent.context).inflate(R.layout.item_message, parent, false)
```
luciofm, line 16: We should lazy initialize a `inflater` field instead of calling  `LayoutInflater.from(parent.context)` on every `onCreateViewHolder()` call

app/src/main/java/chat/rocket/android/app/DateTimeHelper.kt
philipbrito: Some [java.time](https://docs.oracle.com/javase/8/docs/api/java/time/package-summary.html) APIs are available only in the SDK 23+ :confused:

Yeah! We are already [using](https://github.com/filipedelimabrito/Rocket.Chat.Android/blob/f7f0277fd0a9bcdd6eb4c3406f420537c9d4344e/app/src/main/java/chat/rocket/android/app/DateTimeHelper.kt#L4) the ThreeTenABP super cool lib! ❤️

app/src/main/java/chat/rocket/android/app/chatlist/ChatListAdapter.kt
philipbrito: Thanks for advice! When we have the final model we'll do that, for sure! Thanks!

PR #655, opened 2017-12-23
[IMPROVEMENT] Adds error messages to login edit texts

### LINKED ISSUE #640: [IMPROVEMENT] No indication of error when sign in button is pressed without entering username in register screen
- Rocket.Chat.Android app version: 1.0.33-DEBUG
- Device model (or emulator): HTC Desire 628(Physical device)
- Steps to reproduce:
1. Enter the server url, which opens the register screen.
2. On pressing the "sign in" button without entering the username, there is no indication like an error message below the edit text.

This is a bad UI/UX practice and I feel that the following changes can be made to improve this:
1. Keeping the sign in button disabled until username and password are entered.
2. Another way could be to display an error message at the bottom of the edit texts in case button is pressed when username/password field is empty.

The following gif demonstrates this bug:

sign_in_error

aniketsingh03:
@RocketChat/android

Closes #640 

This  pull request adds error messages to the login screen edit texts. This will be a minor improvement in the ux of the app.

### DIFF
--- a/app/src/main/java/chat/rocket/android/fragment/server_config/LoginFragment.kt
+++ b/app/src/main/java/chat/rocket/android/fragment/server_config/LoginFragment.kt
@@ -5,9 +5,7 @@ import android.support.constraint.ConstraintLayout
 import android.support.design.widget.Snackbar
 import android.support.design.widget.TextInputLayout
 import android.support.v4.app.Fragment
-import android.text.Editable
 import android.text.TextUtils
-import android.text.TextWatcher
 import android.view.View
 import android.widget.Button
 import android.widget.TextView
@@ -18,7 +16,9 @@ import chat.rocket.android.log.RCLog
 import chat.rocket.core.models.LoginServiceConfiguration
 import chat.rocket.persistence.realm.repositories.RealmLoginServiceConfigurationRepository
 import chat.rocket.persistence.realm.repositories.RealmPublicSettingRepository
+import com.jakewharton.rxbinding2.widget.RxTextView
 import java.util.*
+
 
 /**
  * Login screen.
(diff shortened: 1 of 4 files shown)


app/src/main/java/chat/rocket/android/fragment/server_config/LoginFragment.kt
```diff
@@ -48,6 +54,38 @@ class LoginFragment : AbstractServerConfigFragment(), LoginContract.View {
         val btnUserRegistration = rootView.findViewById<Button>(R.id.btn_user_registration)
         txtUsername = rootView.findViewById(R.id.editor_username)
         txtPasswd = rootView.findViewById(R.id.editor_passwd)
+        textInputUsername = rootView.findViewById(R.id.text_input_username)
+        textInputPassword = rootView.findViewById(R.id.text_input_passwd)
+
+        //setting text change listeners to username and password edit texts
+        txtUsername.addTextChangedListener(object :TextWatcher
```
philipbrito: Thanks @aniketsingh03 for this PR!

In the future I'd go to recommend you the RxBinding. Your code isn't wrong, but you can avoid the boilerplate. 

Just a friendly recommendation.

app/src/main/java/chat/rocket/android/fragment/server_config/LoginFragment.kt
```diff
@@ -48,6 +54,38 @@ class LoginFragment : AbstractServerConfigFragment(), LoginContract.View {
...
+
+            override fun beforeTextChanged(p0: CharSequence?, p1: Int, p2: Int, p3: Int) {
+            }
+
+            override fun onTextChanged(p0: CharSequence?, p1: Int, p2: Int, p3: Int) {
+                if(!TextUtils.isEmpty(txtPasswd.text.toString()) && textInputPassword.isErrorEnabled)
+                    textInputPassword.setErrorEnabled(false)
+            }
+        })
+
```
leonardoaramaki: I know the IDE names those parameters p0, p1, p2, and son on automatically but when we override some method or anonymous classes, I think we should avoid these defaults and properly naming each one of the parameters.

app/src/main/java/chat/rocket/android/fragment/server_config/LoginFragment.kt
aniketsingh03: Yeah I'll implement it. Thanks for the advice @filipedelimabrito !

aniketsingh03:
@leonardoaramaki @filipedelimabrito I have implemented rxbinders, you can check and merge this.

PR #666, opened 2017-12-26
[NEW] Show list of chats using the SDK (pure online)

### LINKED ISSUE #604: [NEW] Show list of chats using the SDK (pure online)

philipbrito:
Closes #604 

**TODO:**

- [x] Get room name
- [x] Get unread count
- [x] Get room avatar (room initials)
- [x] Get last message (if allowed by server settings -> simple check if content isn't null)
- [x] Get last message timestamp (if allowed by server settings -> simple check if content isn't null)
- [ ] Get system message (e.g  _sent an attachment_) -> Blocked by SDK


app/src/main/java/chat/rocket/android/authentication/login/presentation/LoginView.kt
```diff
@@ -1,7 +1,66 @@
...
+     * Shows the login by Gitlab view.
+     */
+    fun enableLoginByGitlab()
+
+    /**
+     * Shows the sign up view if the server settings allow the new users registration.
+     *
+     * @param show True to show the sign up view, false otherwise.
+     */
+    fun shouldShowSignUpView(show: Boolean)
```
luciofm: Same as above.

app/src/main/java/chat/rocket/android/authentication/login/presentation/LoginView.kt
```diff
@@ -1,7 +1,66 @@
...
+    /**
+     * Shows the oauth view if the server settings allow the login via social accounts.
+     *
+     * REMARK: we must show at maximum *three* social accounts views ([enableLoginByFacebook], [enableLoginByGithub], [enableLoginByGoogle],
+     * [enableLoginByLinkedin], [enableLoginByMeteor], [enableLoginByTwitter] or [enableLoginByGitlab]) for the oauth view.
+     * If the possibility of login via social accounts exceeds 3 different ways we should set up the FAB ([setupFabListener]) to show the remaining view(s).
+     *
+     * @param show True to show the oauth view, false otherwise.
+     */
+    fun shouldShowOauthView(show: Boolean)
```
luciofm: functions named shouldSomething denote a question, like you are asking if we should show something, and not an action... I'd rather prefer 2 methods, show/hideOauthView

app/src/main/java/chat/rocket/android/authentication/login/presentation/LoginPresenter.kt
```diff
@@ -1,51 +1,59 @@
...
-    }
-
-    fun authenticate(username: String, password: String) {
-        // TODO - validate input
-
-        launchUI(strategy) {
-            view.showLoading()
-            try {
-                val token = client.login(username, password)
+    fun authenticate(usernameOrEmailEditText: EditText, passwordEditText: EditText) {
```
luciofm: The presenter should never have access to Android views (widgets), ideally the presenter will never touch any Android specific code.

app/src/main/java/chat/rocket/android/authentication/server/presentation/ServerPresenter.kt
```diff
@@ -1,13 +1,32 @@
...
 
 class ServerPresenter @Inject constructor(private val view: ServerView,
+                                          private val strategy: CancelStrategy,
                                           private val navigator: AuthenticationNavigator) {
+    @Inject lateinit var client: RocketChatClient
 
-    fun login(server: String) {
-        // TODO - validate server URL and get server settings and info before going to Login screen
-        navigator.toLogin(server)
+    fun connect(context: Context, server: String) {
```
luciofm: Again, the presenter should never have access to a Context...

app/src/main/java/chat/rocket/android/authentication/signup/presentation/SignupPresenter.kt
```diff
@@ -1,52 +1,62 @@
...
-
-                navigator.toChatList()
-            } catch (ex: RocketChatException) {
-                view.onSignupError(ex.message)
-            } finally {
-                view.hideLoading()
+                                          private val navigator: AuthenticationNavigator) {
+    @Inject lateinit var client: RocketChatClient
+
+    fun signup(nameEditText: EditText, usernameEditText: EditText, passwordEditText: EditText, emailEditText: EditText) {
```
luciofm: Same here, the presenter must be isolated from Android framework code.

app/src/main/java/chat/rocket/android/dagger/module/AppModule.kt
```diff
@@ -5,18 +5,35 @@ import android.arch.persistence.room.Room
...
 import okhttp3.OkHttpClient
 import okhttp3.logging.HttpLoggingInterceptor
 import javax.inject.Singleton
 
 @Module
 class AppModule {
 
+    @Provides
+    @Singleton
+    fun provideRocketChatClient(okHttpClient: OkHttpClient, repository: AuthTokenRepository, logger: PlatformLogger): RocketChatClient {
```
luciofm: We can't provide the RocketChatClient at the application level because we need the server URL when creating an instance...

The idea is to create a RocketChatClientProvider that could cache clients for different servers (for some time, so we don't need to recreate one every time we change the server)

dependencies.gradle
```diff
@@ -2,81 +2,84 @@ import org.gradle.internal.jvm.Jvm
 
 ext {
     versions = [
-        java              : JavaVersion.VERSION_1_8,
```
luciofm: Try to avoid this kind of changes... It is hard to see if anything really changed or if it is just indentation...

luciofm (changes requested):

app/src/main/java/chat/rocket/android/authentication/login/presentation/LoginPresenter.kt
philipbrito: I agree that the presenter should never have access/dependencies to the Android API's. The presenter should be easily testable. I'll change it.

app/src/main/java/chat/rocket/android/dagger/module/AppModule.kt
philipbrito: I added the provider because it is better than creating a `RocketChatClient` in each presenter as before. About the server url I think that we should don't force the user to specify a specific server when creating a `RocketChatClient` (provide the `AuthTokenRepository` and `PlatformLogger` is another thing that isn't good IMHO). Think about others SDK like AWS or Firebase: it request only the user ID, regions...
I believe that having something like `rocketchatclient.connect("server_url")` would be nice.

app/src/main/java/chat/rocket/android/app/DateTimeHelper.kt
```diff
@@ -36,18 +34,33 @@ object DateTimeHelper {
...
     }
 
-    private fun formatDate(localDate: LocalDate): String {
+    /**
+     * Returns a [LocalDateTime] from a [Long].
+     *
+     * @param long The [Long]
+     * @return The [LocalDateTime] from a [Long].
+     */
+    fun getLocalDateTime(long: Long): LocalDateTime {
```
luciofm, line 53: Why not use extension functions instead of Helper classes?
`fun Long.toLocalDateTime(): LocalDateTime`

`val date = timestamp.toLocalDateTime()`

We get a more idiomatic code and less Helper and Utils classes.

app/src/main/java/chat/rocket/android/app/DrawableHelper.kt
```diff
@@ -37,7 +44,7 @@ object DrawableHelper {
      * @see wrapDrawables
      * @see tintDrawable
      */
-    fun wrapDrawable(drawable: Drawable) = DrawableCompat.wrap(drawable)
+    fun wrapDrawable(drawable: Drawable): Drawable = DrawableCompat.wrap(drawable)
```
luciofm, line 47: I think most of our Helper classes could be extension functions
```
fun Drawable.wrap() = DrawableCompat.wrap(this)
fun Context.drawableForId(id: Int) = this.resources.getDrawable(id, null)
fun Context.getColorCompat(resId: Int) = ContextCompat.getColor(this, resId)
fun Array<Drawable>.tint(context: Context, resId: Int) {
    for (drawable in this) {
        DrawableCompat.setTint(drawable, context.getColorCompat(resId))
    }
}
```

Just some samples...

app/src/main/java/chat/rocket/android/authentication/login/presentation/LoginPresenter.kt
```diff
@@ -1,51 +1,56 @@
...
+                            Timber.d("Created token: $token")
+                            navigator.toChatList()
+                        } catch (rocketChatException: RocketChatException) {
+                            when (rocketChatException) {
+                                is RocketChatTwoFactorException -> {
+                                    navigator.toTwoFA(navigator.currentServer!!, usernameOrEmail, password)
+                                }
+                                else -> {
+                                    val errorMessage = rocketChatException.message
+                                    if (errorMessage != null) {
```
luciofm: If the message is null, we should show a generic error

app/src/main/java/chat/rocket/android/authentication/server/ui/ServerFragment.kt
```diff
@@ -5,44 +5,71 @@ import android.support.v4.app.Fragment
...
-                              container: ViewGroup?,
-                              savedInstanceState: Bundle?): View? =
-            inflater.inflate(R.layout.fragment_authentication_server, container, false)
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? = inflater.inflate(R.layout.fragment_authentication_server, container, false)
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
+        relative_layout.viewTreeObserver.addOnGlobalLayoutListener(layoutListener)
+
+        activity?.applicationContext?.apply {
```
luciofm: no need to use the `applicationContext` here, `activity?.apply` is enough...

app/src/main/java/chat/rocket/android/helper/AnimationHelper.kt
```diff
@@ -0,0 +1,52 @@
...
+import android.animation.AnimatorListenerAdapter
+import android.animation.AnimatorSet
+import android.animation.ObjectAnimator
+import android.content.Context
+import android.os.Build
+import android.os.VibrationEffect
+import android.os.Vibrator
+import android.view.View
+
+object AnimationHelper {
```
luciofm: Again, lets avoid Helper classes and use extensions...

philipbrito: I don't like populating our code with a lot of extensions functions. This class can easily decompiled to a java version, allowing future uses in non-Kotlin projects. Also, if the idea is to create lots of files to keep the extensions functions, what is the diference of having classes like that? Unless we have a big file to keep our extensions (that wouldn't be good).

Always recommended: https://www.philosophicalhacker.com/post/how-to-abuse-kotlin-extension-functions/

app/src/main/java/chat/rocket/android/app/DateTimeHelper.kt
philipbrito: Keep an eye on my comment below... The same for it...

app/src/main/java/chat/rocket/android/app/DrawableHelper.kt
philipbrito: It is a good case to be an extension function. I'll add a TODO. Thanks!

PR #680, opened 2018-01-05
[NEW] Integrated GCM and added notifications support

leonardoaramaki:
@RocketChat/android 

This  PR depends on https://github.com/RocketChat/Rocket.Chat.Kotlin.SDK/pull/33 been merged.


rafaelks:
Looks like the PR has conflicts @leonardoaramaki.

app/src/main/java/chat/rocket/android/authentication/login/presentation/LoginPresenter.kt
```diff
@@ -53,4 +56,15 @@ class LoginPresenter @Inject constructor(private val view: LoginView,
     fun signup() {
         navigator.toSignUp(navigator.currentServer!!)
     }
+
+    private suspend fun registerPushToken() {
+        // TODO: put it on constructor
```
rafaelks: This code could be more generic? I'm sure we'll have to call in different activities later.

app/src/main/java/chat/rocket/android/authentication/twofactor/presentation/TwoFAPresenter.kt
```diff
@@ -47,4 +50,15 @@ class TwoFAPresenter @Inject constructor(private val view: TwoFAView,
     fun signup() {
         navigator.toSignUp(navigator.currentServer!!)
     }
+
+    private suspend fun registerPushToken() {
+        // TODO: put it on constructor
```
rafaelks: There you go... it's already being called twice? This code could definitely be more generic and don't repeated.

app/src/main/java/chat/rocket/android/authentication/login/presentation/LoginPresenter.kt
leonardoaramaki: I agree this has to be discussed better later since we're using coroutines here we don't have callbacks.

app/src/main/java/chat/rocket/android/authentication/login/presentation/LoginPresenter.kt
```diff
@@ -56,4 +61,15 @@ class LoginPresenter @Inject constructor(private val view: LoginView,
     fun signup() {
         navigator.toSignUp()
     }
+
+    private suspend fun registerPushToken() {
+        // TODO: put it on constructor
+        val localRepository: LocalRepository = SharedPreferencesRepository(navigator.activity)
```
luciofm: You should inject the Repository into the Presenter, and the (application)Context can be injected into the repository.

The presenter should never directly know or access any context, Activity or Application

app/src/main/java/chat/rocket/android/authentication/twofactor/presentation/TwoFAPresenter.kt
```diff
@@ -50,4 +54,15 @@ class TwoFAPresenter @Inject constructor(private val view: TwoFAView,
     fun signup() {
         navigator.toSignUp()
     }
+
+    private suspend fun registerPushToken() {
+        // TODO: put it on constructor
+        val localRepository: LocalRepository = SharedPreferencesRepository(navigator.activity)
```
luciofm: Same as on LoginPresenter, also I would create an Interactor for this, so you can inject the same interactor on both presenters and just call a single method on the Interactor.

app/src/main/java/chat/rocket/android/infrastructure/SharedPreferencesRepository.kt
```diff
@@ -0,0 +1,21 @@
+package chat.rocket.android.infrastructure
+
+import android.content.Context
+import android.content.SharedPreferences
+
+class SharedPreferencesRepository(context: Context) : LocalRepository {
```
luciofm: on `AppModule` add a 
```
@Inject
fun provideSharedPreferenceRepository(preferences: SharedPreferences): LocalRepository {
    return  SharedPreferencesRepository(preferences)
}
```

and
```
class SharedPreferencesRepository(private val preferences: SharedPreferences) : LocalRepository {
```
  
So you can inject a `LocalRepository` into Interactors

app/src/main/java/chat/rocket/android/push/PushManager.kt
```diff
@@ -0,0 +1,604 @@
...
+        val style = data["style"] as String?
+        val summaryText = data["summaryText"] as String?
+        val count = data["count"] as String?
+        val title = data["title"] as String?
+
+        if (ejson == null || message == null || title == null) {
+            return
+        }
+
+        val lastPushMessage = PushMessage(title, message, image, ejson, count, notId, summaryText, style)
```
luciofm, line 72: why lastPushMessage and not just pushMessage? With that name seems that you are dealing with an old/previous push message

app/src/main/java/chat/rocket/android/push/di/PushModule.kt
```diff
@@ -0,0 +1,64 @@
...
+import chat.rocket.core.TokenRepository
+import dagger.Module
+import dagger.Provides
+import okhttp3.OkHttpClient
+import okhttp3.logging.HttpLoggingInterceptor
+import javax.inject.Singleton
+
+@Module
+@Singleton
+class PushModule(val context: FirebaseTokenService) {
```
luciofm: Do we need a PushModule, can't it be on the AppModule?

app/src/main/java/chat/rocket/android/authentication/login/presentation/LoginPresenter.kt
leonardoaramaki: Actually, we should make RocketChatApplication a singleton. But the presenter here is not aware of the context, since we get it from the Navigator which is injected already. Ideally though, we should have the appContext here

luciofm: it is already, just as a `Context`
```
    @Provides
    @Singleton
    fun provideContext(application: Application): Context {
        return application
    }
```

luciofm: But either way, the presenter should never directly see or touch the Context...

luciofm (approved):

PR #682, opened 2018-01-05
[NEW] Retrieve public settings on login screens

luciofm:


app/src/main/java/chat/rocket/android/authentication/login/presentation/LoginPresenter.kt
```diff
@@ -3,20 +3,79 @@ package chat.rocket.android.authentication.login.presentation
...
+        }
+
+        view.showSignUpView(settings.registrationEnabled())
+        view.showOauthView(hasSocial)
+    }
 
     fun authenticate(usernameOrEmail: String, password: String) {
+        val server = serverInteractor.get()
         when {
+            server == null -> {
```
philipbrito, line 76: How does the user to come to login screen before the server screen? :thinking:  At this point the server should be non-null, Am I right?

app/src/main/java/chat/rocket/android/authentication/signup/presentation/SignupPresenter.kt
```diff
@@ -3,20 +3,26 @@ package chat.rocket.android.authentication.signup.presentation
...
-                                          private val navigator: AuthenticationNavigator) {
-    @Inject lateinit var client: RocketChatClient
+                                          private val navigator: AuthenticationNavigator,
+                                          private val serverInteractor: GetCurrentServerInteractor,
+                                          private val factory: RocketChatClientFactory) {
 
     fun signup(name: String, username: String, password: String, email: String) {
+        val server = serverInteractor.get()
         when {
+            server == null -> {
```
philipbrito, line 23: Why would we have a null server at this point? When the user comes to Sign Up's screen means that the user has already set up the server before.

philipbrito (changes requested):

app/src/main/java/chat/rocket/android/authentication/login/presentation/LoginPresenter.kt
luciofm: We don't have any verification yet that the server is non-null... We should add a "splashscreen" that would verify if the user has a current server and credentials to this server and go to the chatrooms screens, or to the login screen if not.

PR #688, opened 2018-01-08
[NEW] Support multiple attachments

### LINKED ISSUE #686: [QUESTION] Only Last Attachment is displayed
Hello, im running the latest Version of the server and app. We can not see multiple attachments of one message. Only the last is displayed! Hopefully easy to fix...

Thanks!

Btw: why ist there another Rocket Chat App (reactjs)? This one isnt even finished.

sscholl:
@RocketChat/android

Closes #686

Only the last attachment is shown. This line explains why. If you don't think, that this is correct, please explain why this line is required. Thanks you guys!

### DIFF
--- a/rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/message/RocketChatMessageAttachmentsLayout.java
+++ b/rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/message/RocketChatMessageAttachmentsLayout.java
@@ -79,6 +79,7 @@ public class RocketChatMessageAttachmentsLayout extends LinearLayout {
         }
         this.attachments = attachments;
 
+        removeAllViews();
         for (int i = 0, size = attachments.size(); i < size; i++) {
             appendAttachmentView(attachments.get(i), autoloadImages, true);
         }
--- a/rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/message/RocketChatMessageAttachmentsLayout.java
+++ b/rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/message/RocketChatMessageAttachmentsLayout.java
@@ -89,7 +89,6 @@ public class RocketChatMessageAttachmentsLayout extends LinearLayout {
             return;
         }
 
-        removeAllViews();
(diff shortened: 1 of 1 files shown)


rafaelks:
@sscholl Some of the Android developers may have a better answer, but my understanding is: we're not showing multiple attachments because we don't loop into the attachments objects. This line you pointed, IMHO, is used because we reuse the cell in the list of messages instead of allocating new objects to every new message.

sscholl:
@rafaelks so this remove line is added for performance?

I do see the loop into attachments:

https://github.com/RocketChat/Rocket.Chat.Android/blob/fcc9a2a5ce5508d9d93ee76fbe20a6c6fb4a2454/rocket-chat-android-widgets/src/main/java/chat/rocket/android/widget/message/RocketChatMessageAttachmentsLayout.java#L82-L84
 And "appendAttachmentView" is only invocated at this point.
  
EDIT: The removeAllViews(); was moved in this commit and I think this was wrong:
https://github.com/RocketChat/Rocket.Chat.Android/commit/ab0c5329ea08a544b31539ecb90d154ce8b8fe92?diff=split

rafaelks:
@sscholl Can you confirm that the attachments objects actually have all the objects you want to? Maybe we're not mapping all of them from the JSON.

sscholl:
@rafaelks Yes. I have 1-4 identically generated attachments (all by the same code). Every time, only the last one is shown. However, in normal web view, it's showing.

rafaelks:
@sscholl Removing this line of code worked for you?

sscholl:
@rafaelks I have added the removeAllViews() invocation from commit de3b81edb513aa55d1d43ea8f8600c7bc5f10cd2 again, but before the loop. This is also the only place where it makes sense. I tested it on my phone with my bot, that shows these multiple attachements. (Unfortually, Markdown is not working, but this is another topic)

sscholl:
@rafaelks is this PR ok or do I need to put more work in it?

philipbrito (approved):

sscholl:
@filipedelimabrito thanks a lot!

philipbrito:
Thanks for your contribution!

PR #695, opened 2018-01-12
[NEW] Spotlight (local data)

leonardoaramaki:
@RocketChat/android

Implements the spotlight but searching locally in-memory.


app/src/main/java/chat/rocket/android/chatrooms/ui/ChatRoomsFragment.kt
```diff
@@ -33,9 +33,18 @@ class ChatRoomsFragment : Fragment(), ChatRoomsView {
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
         presenter.chatRooms()
+        floating_search_view.setOnQueryChangeListener { oldQuery, newQuery ->
```
rafaelks, line 36: Shouldn't the `floating_search_view` name be camelCase? 🤔

app/src/main/java/chat/rocket/android/server/domain/GetChatRoomsInteractor.kt
```diff
@@ -0,0 +1,22 @@
...
+
+    suspend fun getByName(url: String, name: String): List<ChatRoom> {
+        val chatRooms = async {
+            val allChatRooms = repository.get(url)
+            if (name.isEmpty()) {
+                return@async allChatRooms
+            }
+            return@async allChatRooms.filter {
+                it.name.contains(name, true)
+            }
```
rafaelks, line 18: Will this one also filter DMs?

app/src/main/java/chat/rocket/android/chatrooms/ui/ChatRoomsFragment.kt
leonardoaramaki: Usually yes. But in this case it's in snake_case because of the Kotlin plugin we use: https://antonioleiva.com/kotlin-android-extensions. What we should agree is to name our view ids in camelCase since we're using this extension.

philipbrito: Stepping in for @leonardoaramaki  the convention for our xml files is right, but because we use Kotlin Android Extension we don't need to use the (boring) findViewById, get the result and assign to a variable (e.g. `floatingSearchView`). We only need to refer the xlm id (`floating_search_view`) and KAE does the rest (find it, cache it...)

app/src/main/java/chat/rocket/android/server/domain/GetChatRoomsInteractor.kt
luciofm: Yes, chatrooms includes all types of "rooms"

rafaelks: 👍

PR #696, opened 2018-01-12
[NEW] Support sending message using SDK

### LINKED ISSUE #611: [NEW] Support sending a message with the SDK
For now, just text message.

philipbrito:
Closes #611 #256


app/build.gradle
```diff
@@ -34,7 +34,7 @@ android {
 }
 
 dependencies {
-    implementation fileTree(dir: 'libs', include: ['*.jar'])
+    implementation fileTree(include: ['*.jar'], dir: 'libs')
```
leonardoaramaki, line 37: What's the difference this swapping makes?

philipbrito: As you know, I had troubles when integrating the SDK into the app :(. After adding the jars as a library through the IDE this dependency declaration has changed.

PR #701, opened 2018-01-15
[NEW] stream room messages

luciofm:


app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomAdapter.kt
```diff
@@ -41,6 +50,15 @@ class ChatRoomAdapter(private val context: Context,
...
+    fun updateItem(index: Int, message: Message) {
+        dataSet[index] = message
+        notifyItemChanged(index)
+    }
+
+    override fun getItemId(position: Int): Long {
+        return dataSet[position].id.hashCode().toLong()
+    }
+
     inner class ViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
```
leonardoaramaki, line 62: Why set `ViewHolder` class as `inner`? You can access a context with `itemView.context`

luciofm: @filipedelimabrito ?

philipbrito: @leonardoaramaki  We don't need to get the context of the itemView (`image_avatar`, `image_unknown_avatar`, `text_user_name`, `text_message_time` or `text_content`). We're just passing the context from the activity https://github.com/RocketChat/Rocket.Chat.Android/blob/9d59c945c5f14ce9637cb282ab3d3fef9ce2a558/app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt#L79 

Pls, read it (I think that it is a valid reference even for this response): https://stackoverflow.com/a/12139064/4744263

c/c @luciofm

philipbrito (approved):

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomAdapter.kt
leonardoaramaki: I know... But by getting the context from itemView you get the Activity's context for free without needing to stick the ViewHolders classes with the adapter's

philipbrito: In your (prefered?) way you will get the activity context two times (calling it in `ChatRoomFragment.kt` and  `ChatRoomAdapter.kt`)...

Anyway, I think that it is just a way of getting things working. 
Hope that you can understand the arguments.

leonardoaramaki: Why preferred? It's just usually the most correct way to do it since you don't need to hold a strong 
 reference to the context on the adapter and neither the ViewHolder. We don't need to get the context two times. Actually we don't need to hold it anywhere, just use it on the adapter the same way you use it on the ViewHolder.

philipbrito: Where did you found that it is the most correctly way to do it? Could you provide me a source/link/etc?
Even the View Holder design pattner doesn't justify such argumentation.

Well, if you check the referenced app by Google (https://github.com/google/iosched/blob/e4364eec0d8f358215aee30cce2b4836ec169b5a/lib/src/main/java/com/google/samples/apps/iosched/schedule/SessionsFilterAdapter.java#L62) you will find that it isn't done with that way.

Don't panic! I hope that you can understand the arguments.

leonardoaramaki: Sure, I understood yours. I'm just trying to show you a way that you don't have to keep any Context references on any adapter. 
Look at your own reference from Google. Even on your example they don't keep a reference to it. They just get it and uset it, just like I said.

Here is ours:
https://github.com/RocketChat/Rocket.Chat.Android/blob/9d59c945c5f14ce9637cb282ab3d3fef9ce2a558/app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomAdapter.kt#L21

You see the difference?

luciofm: Ideally inner classes should be static (on Java) so they don't hold a reference to the outer class...

This is problematic on Android because of the lifecycle, if the inner class lives longer than the outer class, it will leak the outer class as long as it is alive.

PR #705, opened 2018-01-16
[NEW] Save login information and server settings

luciofm:
Saving server settings to SharedPreferences for now... Will move to a proper  database in the future to support settings streaming.


app/src/main/java/chat/rocket/android/authentication/domain/CurrentTokenRepository.kt
```diff
@@ -0,0 +1,11 @@
+package chat.rocket.android.authentication.domain
+
+/**
```
philipbrito: @luciofm Could you remove it?

app/src/main/java/chat/rocket/android/authentication/domain/CurrentTokenRepository.kt
```diff
@@ -0,0 +1,11 @@
...
+ * Created by luciofm on 1/15/18.
+ */
+
+interface CurrentTokenRepository {
```
philipbrito: From the function signature we are saving and getting the `server` so could you change  `CurrentTokenRepository` to `CurrentServerRepository`? What do you think?

app/src/main/java/chat/rocket/android/authentication/infraestructure/SharedPreferencesMultiServerTokenRepository.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    override fun save(server: String, token: TokenModel) {
+        val adapter = moshi.adapter<TokenModel>(TokenModel::class.java)
+
+        repository.save("$TOKEN_KEY$server", adapter.toJson(token))
+    }
+
+}
+
+const val TOKEN_KEY = "token_"
```
philipbrito, line 33: This is related to the "customer's taste", but once we are defining the class properties on the top we could put our `const` on the top of the class too.

A little of history: usually a developer first looks at class behaviour's implementation (member functions or methods). In some projects we can see the member variables at the bottom of the class and I really like this approach (but this is unusual on Android app development) - making the class easy to read (like you did here).

app/src/main/java/chat/rocket/android/dagger/module/AppModule.kt
```diff
@@ -88,8 +91,8 @@ class AppModule {
 
     @Provides
     @Singleton
-    fun provideAuthTokenRepository(): AuthTokenRepository {
-        return AuthTokenRepository()
+    fun provideAuthTokenRepository(): TokenRepository {
+        return MemoryTokenRepository()
```
philipbrito: Now that we are returning the `MemoryTokenRepository` we could rename the function to `provideMemoryTokenRepository()`.

app/src/main/java/chat/rocket/android/server/infraestructure/SharedPreferencesSettingsRepository.kt
```diff
@@ -0,0 +1,26 @@
...
+        val settings = localRespository.get("$SETTINGS_KEY$url")
+        settings?.let {
+            return adapter.fromJson(it)
+        }
+
+        return null
+    }
+}
+
+const val SETTINGS_KEY = "settings_"
```
philipbrito: You could put  the `SETTINGS_KEY` at the top of the `SharedPreferencesSettingsRepository` class.

app/src/main/java/chat/rocket/android/server/infraestructure/SharedPrefsCurrentServerRepository.kt
```diff
@@ -4,14 +4,15 @@ import android.content.SharedPreferences
...
     override fun get(): String? {
-        return preferences.getString(CURRENT_SERVER, null)
+        return preferences.getString(CURRENT_SERVER_KEY, null)
     }
 
-}
\ No newline at end of file
+}
+
+const val CURRENT_SERVER_KEY = "current_server"
```
philipbrito: Here too.

app/src/main/res/drawable/splash_screen.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <item>
+        <color android:color="@color/colorPrimary"/>
+    </item>
+
+    <!--<item>
```
philipbrito, line 8: That's great that our splash is only a `bg` with our primary color. I wouldn't think it is good to include the bitmap... so you can remove this commented element.

Why I don't like having a splash (at least with an image)? 
A: we can't see it in Telegram, Whats App and Slack. But maybe @rafaelks can tell us if our mobile app will have a splash screen (with image and so on...).

Old but gold: 
https://cyrilmottier.com/2012/05/03/splash-screens-are-evil-dont-use-them/
https://www.bignerdranch.com/blog/splash-screens-the-right-way/

philipbrito (changes requested):

app/src/main/java/chat/rocket/android/dagger/module/AppModule.kt
luciofm: Actually I think this one is right and we should rename the others... We are hiding the implementation, so we provide a `TokenRepository`, doesn't matter for the user if it is a `MemoryTokenRepository` or a `DatabaseTokenRepository`

app/src/main/java/chat/rocket/android/authentication/domain/CurrentTokenRepository.kt
luciofm: Will remove this interface, it is not used...

app/src/main/res/drawable/splash_screen.xml
luciofm: https://material.io/guidelines/patterns/launch-screens.html

app/src/main/java/chat/rocket/android/server/infraestructure/SharedPreferencesSettingsRepository.kt
luciofm: Actually I think this belongs to a companion object... It is intended to be used just by this class...

Either way, I think it belongs to the bottom of the class.

Quoting the Kotlin Coding Convetions doc

>Class layout
Generally, the contents of a class is sorted in the following order:
-Property declarations and initializer blocks
-Secondary constructors
-Method declarations
-Companion object

app/src/main/java/chat/rocket/android/authentication/infraestructure/SharedPreferencesMultiServerTokenRepository.kt
luciofm: Constants should be self descriptive, there is no need for them to be at the top of the file. And moving to a companion object I'm following the kotlin coding convetions doc.

app/src/main/res/drawable/splash_screen.xml
philipbrito: Nice example! From the link we can see how Gmail actually does: it shows the branded launch screen at the first opening and after it, when we open the app it is not show anymore. In that way, I believe that you can remove the commented element (it isn't in use).

app/src/main/java/chat/rocket/android/server/infraestructure/SharedPreferencesSettingsRepository.kt
philipbrito: Nice! And we never put our companion objects bellow the method declarations :grin:

app/src/main/java/chat/rocket/android/authentication/infraestructure/SharedPreferencesMultiServerTokenRepository.kt
philipbrito: A good example of putting the `const` at the top is here (when passing extras - not this case, just for reference): https://github.com/RocketChat/Rocket.Chat.Android/blob/develop-2.x/app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomActivity.kt

app/src/main/java/chat/rocket/android/authentication/infraestructure/SharedPreferencesMultiServerTokenRepository.kt
```diff
@@ -0,0 +1,33 @@
...
+@PerActivity
+class SharedPreferencesMultiServerTokenRepository(private val repository: LocalRepository,
+                                                  private val moshi: Moshi
+) : MultiServerTokenRepository {
+
+    override fun get(server: String): TokenModel? {
+        val token = repository.get("$TOKEN_KEY$server")
+        val adapter = moshi.adapter<TokenModel>(TokenModel::class.java)
+
+        token?.let {
```
philipbrito, line 18: Forgot to say (in a past PR): thanks!
This is better than checking if the token isn't null before passing it in `.fromJson(token)`. And we can use the extension `ifNull`.
Thanks for the knowledge.
:clap: :clap:

app/src/main/java/chat/rocket/android/authentication/infraestructure/SharedPreferencesMultiServerTokenRepository.kt
luciofm: I can't see any value having the companion objects at the top of the file, I think we should follow the Coding Conventions doc for Kotlin -> https://kotlinlang.org/docs/reference/coding-conventions.html, more specifically the [Class Layout](https://kotlinlang.org/docs/reference/coding-conventions.html#class-layout)

A companion object is just like a helper object, or a collection of helper constants and small functions, that's why it belongs to the bottom of the class

philipbrito: I followed this nice article by the Eugenio @Takhion Marletti (probably you meet him since he is a GDE member like you):

https://medium.com/@workingkills/you-wont-believe-this-one-weird-trick-to-handle-android-intent-extras-with-kotlin-845ecf09e0e9

There is a nice argumentation there. We could discuss it later.

philipbrito: Oh sorry, the article that I would like to share is: https://medium.com/@passsy/starting-activities-with-kotlin-my-journey-8b7307f1e460

This is also another nice article by the GDE (for Flutter :P)  Pascal @passsy Welsch

PR #706, opened 2018-01-17
[NEW] Honor Use Real Name setting

### LINKED ISSUE #363: [NEW] Honor setting to show first name / last name

leonardoaramaki:
@RocketChat/android

Closes #363

### DIFF
--- a/app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomView.kt
+++ b/app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomView.kt
@@ -3,6 +3,7 @@ package chat.rocket.android.chatroom.presentation
 import chat.rocket.android.core.behaviours.LoadingView
 import chat.rocket.android.core.behaviours.MessageView
 import chat.rocket.core.model.Message
+import chat.rocket.core.model.Value
 
 interface ChatRoomView : LoadingView, MessageView {
 
@@ -11,8 +12,9 @@ interface ChatRoomView : LoadingView, MessageView {
      *
      * @param dataSet The data set to show.
      * @param serverUrl The server URL.
+     * @param settings The server settings.
      */
-    fun showMessages(dataSet: List<Message>, serverUrl: String)
+    fun showMessages(dataSet: List<Message>, serverUrl: String, settings: Map<String, Value<Any>>?)
 
     /**
      * Send a message to a chat room.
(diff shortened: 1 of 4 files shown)


app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomAdapter.kt
```diff
@@ -74,12 +77,20 @@ class ChatRoomAdapter(private val context: Context,
             imageUnknownAvatar.setVisibility(true)
         }
 
-        private fun bindUserName(message: Message, textView: TextView) = message.sender?.username.let {
-            textView.textContent = it.toString()
-        }.ifNull {
-            textView.textContent = context.getString(R.string.msg_unknown)
-        }
+        private fun bindUserName(message: Message, textView: TextView) {
```
luciofm, line 80: For now we can live with this logic inside the Adapter, but in the future, when we have local persistence, I'd like to use a mapper to deliver a  "MessageViewModel" to the adapter that is really ready to be "rendered", All the information should be ready, including spannables with links, emojis, etc...

app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomView.kt
```diff
@@ -11,8 +12,9 @@ interface ChatRoomView : LoadingView, MessageView {
      *
      * @param dataSet The data set to show.
      * @param serverUrl The server URL.
+     * @param settings The server settings.
      */
-    fun showMessages(dataSet: List<Message>, serverUrl: String)
+    fun showMessages(dataSet: List<Message>, serverUrl: String, settings: Map<String, Value<Any>>?)
```
luciofm, line 17: Why not pass only the setting value, instead of all settings?

leonardoaramaki: I thought on passing all the settings because we may have other settings the app could use later to customize the messages.
Actually, I was thinking about creating a sort of view model class to accomplish that, which I think would be better.

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomAdapter.kt
leonardoaramaki: Agreed.

app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomView.kt
luciofm: Yeah, if you would like to create one already, it would be better. It is nice because we can do all the conversions, formatting, etc on the background, freeing the main thread

leonardoaramaki:
Closing this one as this feature is implemented on the other PR -> [https://github.com/RocketChat/Rocket.Chat.Android/pull/711](url)

Issue #699, opened 2018-01-12
[NEW] Support custom type messages (user joined, message deleted...)


rafaelks:
(blocked by https://github.com/RocketChat/Rocket.Chat.Kotlin.SDK/issues/40)

PR #711, opened 2018-01-23
[NEW] Render system messages

### LINKED ISSUE #363: [NEW] Honor setting to show first name / last name

leonardoaramaki:
@RocketChat/android

Closes #363 and #699 

System messages are now rendered in the chat accordingly. Since I implemented a view model class to represent the Message entity here I thought those changes are related to the #363 so I've merged both issues into this one.

### DIFF
--- a/app/src/main/java/chat/rocket/android/app/RocketChatApplication.kt
+++ b/app/src/main/java/chat/rocket/android/app/RocketChatApplication.kt
@@ -25,6 +25,10 @@ class RocketChatApplication : Application(), HasActivityInjector, HasServiceInje
     @Inject
     lateinit var serviceDispatchingAndroidInjector: DispatchingAndroidInjector<Service>
 
+    companion object {
+        lateinit var instance: RocketChatApplication
+    }
+
     override fun onCreate() {
         super.onCreate()
 
@@ -34,6 +38,7 @@ class RocketChatApplication : Application(), HasActivityInjector, HasServiceInje
 
         setupFresco()
         setupTimber()
+        instance = this
     }
 
     private fun setupFresco() {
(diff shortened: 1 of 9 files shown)


app/src/main/java/chat/rocket/android/app/RocketChatApplication.kt
```diff
@@ -34,6 +38,7 @@ class RocketChatApplication : Application(), HasActivityInjector, HasServiceInje
 
         setupFresco()
         setupTimber()
+        instance = this
```
luciofm, line 41: You can inject it with dagger... not that makes much difference, but...

app/src/main/java/chat/rocket/android/chatroom/viewmodel/MessageViewModelMapper.kt
```diff
@@ -0,0 +1,17 @@
+package chat.rocket.android.chatroom.viewmodel
+
+import chat.rocket.core.model.Message
+import chat.rocket.core.model.Value
+
+object MessageViewModelMapper {
+
+    suspend fun mapToViewModel(message: Message, settings: Map<String, Value<Any>>?) = MessageViewModel(message, settings)
+
+    suspend fun mapToViewModelList(messageList: List<Message>, settings: Map<String, Value<Any>>?): List<MessageViewModel> {
```
luciofm, line 10: maybe with a `withContext(CommonPool) { }` so we can guarantee that it always run on background...

Coroutines are really light and cheap.

PR #724, opened 2018-01-26
[NEW] Show attachments

luciofm:
There is a lot of stuff that I want to improve here...

The ViewModel will be just a data class, the mapper will do all the heavy lifting (we can inject settings, token, etc into the mapper)
Separate ViewModels and ViewHolders for message, image, video, audio, url preview, etc.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -12,8 +12,8 @@ android {
         applicationId "chat.rocket.android"
         minSdkVersion 21
         targetSdkVersion versions.targetSdk
-        versionCode 1006
-        versionName "2.0.0-dev5"
+        versionCode 1007
+        versionName "2.0.0-dev5.1"
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
         multiDexEnabled true
     }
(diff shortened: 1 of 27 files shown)


app/build.gradle
```diff
@@ -11,8 +12,8 @@ android {
         applicationId "chat.rocket.android"
         minSdkVersion 21
         targetSdkVersion versions.targetSdk
-        versionCode 1001
-        versionName "2.0.0-dev1"
+        versionCode 1005
```
leonardoaramaki: What's the criteria used to update the versionCode?

app/src/main/java/chat/rocket/android/chatroom/viewmodel/MessageViewModel.kt
```diff
@@ -41,7 +69,7 @@ data class MessageViewModel(private val message: Message,
         val username = message.sender?.username
         val realName = message.sender?.name
         val senderName = if (useRealName) realName else username
-        return if (senderName == null) username.toString() else senderName.toString()
+        return senderName ?: username.toString()
```
leonardoaramaki, line 72: :+1:

app/build.gradle
luciofm: For now, each version sent get a versionCode bump for update the apk on the smartphone

Issue #710, opened 2018-01-22
[NEW] Show/update user profile


philipbrito:
Creates a view to show to the user its profile (displaying its email address, real name, username and avatar). Also enables to the user the possibility to update the mentioned fields plus reseting the avatar.

SyamSundarKirubakaran:
@filipedelimabrito I would wish to work on this issue.

philipbrito:
Hi @SyamSundarKirubakaran Currently this issue is almost done. We need only to update/reset the user avatar. Can you work on #752, #753 or #754 ?

SyamSundarKirubakaran:
Ok @filipedelimabrito , I'll look into those.

rafaelks:
This one can be closed already.

PR #725, opened 2018-01-29
[NEW] Show/update user profile

philipbrito:
Closes partly #710 (needs a better UI and update/reset user avatar and password).

Depends on https://github.com/RocketChat/Rocket.Chat.Kotlin.SDK/pull/55


app/src/main/java/chat/rocket/android/authentication/twofactor/presentation/TwoFAPresenter.kt
```diff
@@ -44,9 +44,13 @@ class TwoFAPresenter @Inject constructor(private val view: TwoFAView,
                         try {
                             // The token is saved via the client TokenProvider
                             val token = client.login(usernameOrEmail, password, twoFactorAuthenticationCode)
-                            multiServerRepository.save(server, TokenModel(token.userId, token.authToken))
-                            registerPushToken()
-                            navigator.toChatList()
+                            if (token != null) {
+                                multiServerRepository.save(server, TokenModel(token.userId, token.authToken))
```
leonardoaramaki: I'm not sure but I think if token happens to be null it will fall through to the catch to `RocketChatException`. :thinking:

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomAdapter.kt
```diff
@@ -58,8 +57,8 @@ class ChatRoomAdapter(private val serverUrl: String) : RecyclerView.Adapter<Chat
 
         fun bind(message: MessageViewModel) = with(itemView) {
             bindUserAvatar(message, image_avatar, image_unknown_avatar)
-            text_user_name.text = message.sender
-            text_message_time.text = message.time
+            text_user_name.textContent = message.sender
+            text_message_time.textContent = message.time
```
leonardoaramaki: `textContent` is not suitable here as it expects it expects a String and will coerce to this type clearing all spannables we set.

app/src/main/java/chat/rocket/android/chatroom/viewmodel/MessageViewModel.kt
```diff
@@ -18,14 +18,14 @@ import chat.rocket.core.model.Value
 data class MessageViewModel(private val message: Message,
                             private val settings: Map<String, Value<Any>>?) {
     val id: String = message.id
-    val time: CharSequence
-    val sender: CharSequence
-    val content: CharSequence
+    val time: String
+    val sender: String
+    val content: SpannableString
```
leonardoaramaki: Please, leave them all as `CharSequence` since it mostly expects a Spannable.

app/src/main/java/chat/rocket/android/chatroom/viewmodel/MessageViewModel.kt
```diff
@@ -34,41 +34,40 @@ data class MessageViewModel(private val message: Message,
...
+            else -> getNormalMessage()
         }
-        return contentMessage
     }
 
-    private fun getNormalMessage() = message.message
+    private fun getNormalMessage(): SpannableString = SpannableString(message.message)
 
-    private fun getSystemMessage(content: String): CharSequence {
+    private fun getSystemMessage(content: String): SpannableString {
```
leonardoaramaki: Please, leave it with its supertype `CharSequence` here.

app/src/main/java/chat/rocket/android/chatrooms/ui/ChatRoomsFragment.kt
```diff
@@ -41,45 +40,27 @@ class ChatRoomsFragment : Fragment(), ChatRoomsView {
...
+        setupToolbar()
+        setupRecyclerView()
         presenter.loadChatRooms()
     }
 
-    override fun onCreateOptionsMenu(menu: Menu?, inflater: MenuInflater?) {
-        inflater?.inflate(R.menu.chatrooms_menu, menu)
-        val searchItem = menu?.findItem(R.id.action_search)
+    override fun onCreateOptionsMenu(menu: Menu, inflater: MenuInflater) {
+        super.onCreateOptionsMenu(menu, inflater)
```
leonardoaramaki, line 59: The super class method does nothing so no need to call it.

app/src/main/java/chat/rocket/android/authentication/twofactor/presentation/TwoFAPresenter.kt
philipbrito: With the changes made on the [RestResult](https://github.com/filipedelimabrito/Rocket.Chat.Kotlin.SDK/blob/d71e51b5b566a8d6762c29201fdc3eee52702dcf/core/src/main/kotlin/chat/rocket/core/internal/rest/RestResult.kt) and [RestMultiResult](https://github.com/filipedelimabrito/Rocket.Chat.Kotlin.SDK/blob/d71e51b5b566a8d6762c29201fdc3eee52702dcf/core/src/main/kotlin/chat/rocket/core/internal/rest/RestMultiResult.kt) class we expected null results because `fromJson` can return null.

app/src/main/java/chat/rocket/android/profile/presentation/ProfilePresenter.kt
```diff
@@ -0,0 +1,66 @@
...
+                view.hideLoading()
+            }
+        }
+    }
+
+    fun updateUserProfile(email: String, name: String, username: String) {
+        launchUI(strategy) {
+            view.showLoading()
+            try {
+                val user = client.updateProfile(myselfId, email, name, username)
```
leonardoaramaki, line 47: We need to check it later with @luciofm too, but I do think that we would or at least should fall throw to catch to the `RocketChatException` on all calls.

dependencies.gradle
```diff
@@ -4,7 +4,7 @@ ext {
         compileSdk        : 27,
         targetSdk         : 27,
         buildTools        : '27.0.3',
-        kotlin            : '1.2.10',
+        kotlin            : '1.2.21',
```
leonardoaramaki: :+1:

leonardoaramaki (changes requested):
:)

app/src/main/java/chat/rocket/android/authentication/twofactor/presentation/TwoFAPresenter.kt
leonardoaramaki: Hum. I see. Let's wait for that SDK PR to be fully reviewed then. :smile: 
I think the generics type T? should be T there if possible. I think the best would to have the result or throw.

luciofm: We should never expect null results from the SDK, I was going to comment on those changes... If the SDK doesn't return a value, it should throw...

philipbrito: Are you saying that if the result from two mentioned above class is null (for `result`, `update` and `remove`) we should throw an exception? Is it? I'll make changes there so. 

As soon as you do the review, I'll look into to merge those PRs as quickly as possible to deliver the features..

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomAdapter.kt
luciofm: We should change `textContent` to work with `CharSequence` instead of `String`

app/src/main/java/chat/rocket/android/profile/presentation/ProfilePresenter.kt
luciofm: @leonardoaramaki you are right, we will need to add exception handling code (it can be done with a default handler for the coroutine), but for now, crashing is good because we got to fix the real problem instead of just hiding it.

app/src/main/java/chat/rocket/android/authentication/twofactor/presentation/TwoFAPresenter.kt
luciofm: I commented on the SDK PR, the calls should be nonnullable.

luciofm (changes requested):

app/build.gradle
```diff
@@ -71,35 +71,36 @@ dependencies {
...
-    implementation libraries.floatingSearchView
+    implementation libraries.frescoImageViewer
 
     implementation libraries.androidSvg
 
     implementation libraries.aVLoadingIndicatorView
 
     implementation libraries.textDrawable
 
+    implementation libraries.moshiLazyAdapters
```
luciofm, line 102: where we are using this?

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
```diff
@@ -75,10 +75,10 @@ class ChatRoomFragment : Fragment(), ChatRoomView {
         super.onDestroyView()
     }
 
-    override fun showMessages(dataSet: List<MessageViewModel>, serverUrl: String) {
+    override fun showMessages(dataSet: List<Message>, serverUrl: String) {
```
luciofm: This is really strange, why we are not using the MessageViewModel anymore?

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
```diff
@@ -75,10 +75,10 @@ class ChatRoomFragment : Fragment(), ChatRoomView {
...
         activity?.apply {
             if (recycler_view.adapter == null) {
-                adapter = ChatRoomAdapter(serverUrl)
+                adapter = ChatRoomAdapter(this, serverUrl)
```
luciofm: ChatRoomAdapter doesn't have this constructor...

luciofm (changes requested):

app/build.gradle
philipbrito: Good catch! This was in use with the SDK (from this PR: https://github.com/RocketChat/Rocket.Chat.Kotlin.SDK/pull/55 - related to FallbackEnum). I declared this dependency here because the app was not able to find it by the SDK. Removing it...

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
philipbrito: Something was wrong with the merges... Really strange.. Fixing it manually.

app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomPresenter.kt
```diff
@@ -60,8 +60,8 @@ class ChatRoomPresenter @Inject constructor(private val view: ChatRoomView,
                 ex.message?.let {
                     view.showMessage(it)
                 }.ifNull {
-                    view.showGenericErrorMessage()
-                }
+                            view.showGenericErrorMessage()
```
luciofm: Wrong indentation.

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomAdapter.kt
```diff
@@ -119,8 +119,8 @@ class ChatRoomAdapter(private val serverUrl: String) : RecyclerView.Adapter<Chat
             drawee.setVisible(true)
             imageUnknownAvatar.setVisible(false)
         }.ifNull {
-            drawee.setVisible(false)
-            imageUnknownAvatar.setVisible(true)
-        }
+                    drawee.setVisible(false)
```
luciofm: wrong indentation

app/src/main/java/chat/rocket/android/profile/presentation/ProfilePresenter.kt
```diff
@@ -0,0 +1,61 @@
...
+    private lateinit var myselfId: String
+
+    fun loadUserProfile() {
+        launchUI(strategy) {
+            view.showLoading()
+            try {
+                val myself = client.me()
+                myselfId = myself.id
+                val avatarUrl = UrlHelper.getAvatarUrl(serverUrl, myself.username!!)
+                view.showProfile(avatarUrl, myself.name!!, myself.username!!, myself.emails?.get(0)?.address!!)
```
luciofm, line 30: Don't we need to check all this nullable fields? Or make them nullable on the View interface?

app/src/main/java/chat/rocket/android/util/Extensions.kt
```diff
@@ -35,4 +40,11 @@ var TextView.hintContent: String
     get() = hint.toString()
     set(value) {
         hint = value
-    }
\ No newline at end of file
+    }
+
+fun EditText.getObservable(): Observable<CharSequence> {
```
luciofm, line 45: What do you think about `asObservable`? I think it is a better name.
Or maybe
```
val EditText.observable: Observable<CharSequence>
    get() = RxTextView.textChanges(this)
                  .debounce(100, TimeUnit.MILLISECONDS)
                  .observeOn(AndroidSchedulers.mainThread())
                  .subscribeOn(AndroidSchedulers.mainThread())
```

luciofm (changes requested):

PR #732, opened 2018-02-02
[NEW] Markdown support

### LINKED ISSUE #719: [NEW] Support Markdown on message text

leonardoaramaki:
@RocketChat/android

Closes #719


app/src/main/java/chat/rocket/android/chatroom/viewmodel/MessageViewModelMapper.kt
```diff
@@ -1,16 +1,22 @@
...
 import javax.inject.Inject
 
-class MessageViewModelMapper @Inject constructor(private val context: Context, private val tokenRepository: TokenRepository) {
+class MessageViewModelMapper @Inject constructor(private val context: Context,
+                                                 private val tokenRepository: TokenRepository,
+                                                 private val messageParser: MessageParser,
+                                                 private val messagesRepository: MessagesRepository) {
 
-    suspend fun mapToViewModel(message: Message, settings: Map<String, Value<Any>>?) = MessageViewModel(context, tokenRepository.get(), message, settings)
+    suspend fun mapToViewModel(message: Message, settings: Map<String, Value<Any>>?) =
```
luciofm: lets make it
```kotlin
suspend fun mapToViewModel(message: Message, settings: Map<String, Value<Any>>?) = launch(CoomonPool) {
    MessageViewModel(context, tokenRepository.get(), message, settings, messageParser, messagesRepository)
}
```

app/src/main/java/chat/rocket/android/helper/MessageParser.kt
```diff
@@ -0,0 +1,177 @@
...
+    /**
+     * Render a markdown text message to Spannable.
+     *
+     * @param text The text message containing markdown syntax.
+     * @param quote An optional message to be quoted either by a quote or reply action.
+     * @param urls A list of urls to convert to markdown link syntax.
+     *
+     * @return A Spannable with the parsed markdown.
+     */
+    fun renderMarkdown(text: String, quote: MessageViewModel?, urls: List<Url>): CharSequence {
```
luciofm, line 43: Lets make it suspend and run as `launch(CoomonPool)`

app/src/main/java/chat/rocket/android/helper/MessageParser.kt
```diff
@@ -0,0 +1,177 @@
...
+
+            // pass to super to apply markdown
+            super.visit(blockQuote)
+
+            builder.setSpan(QuoteMarginSpan(context.getDrawable(R.drawable.quote), 10), length, builder.length())
+        }
+    }
+
+    class QuoteMarginSpan(b: Drawable, pad: Int) : LeadingMarginSpan, LineHeightSpan {
+        private var mDrawable: Drawable? = b
```
luciofm: why nullable if the constructor's Drawable is nonnull.
```kotlin
class QuoteMarginSpan(private val drawable, private val pad)
```
And please, no Hungarian notation :pray:

app/src/main/java/chat/rocket/android/util/Extensions.kt
```diff
@@ -36,6 +44,16 @@ var TextView.textContent: String
         text = value
     }
 
+var TextView.content: CharSequence
+    get() = text
+    set(value) {
+        Markwon.unscheduleDrawables(this)
```
luciofm, line 50: Do we need this for all TextViews? Maybe a simple version for TextViews, and another property for markdown texts.

app/src/test/java/chat/rocket/android/MemoryMessagesRepositoryTest.kt
```diff
@@ -0,0 +1,135 @@
...
+import chat.rocket.android.server.infraestructure.MemoryMessagesRepository
+import chat.rocket.core.model.Message
+import chat.rocket.core.model.MessageType
+import org.hamcrest.CoreMatchers.notNullValue
+import org.hamcrest.MatcherAssert.assertThat
+import org.junit.Before
+import org.junit.Test
+import org.hamcrest.CoreMatchers.`is` as isEqualTo
+
+class MemoryMessagesRepositoryTest {
```
luciofm, line 12: Nice, we have a big tech debit on tests

luciofm (changes requested):

app/src/main/java/chat/rocket/android/helper/MessageParser.kt
leonardoaramaki: Nice catch. And sorry it was unintentional. That's and update version of another span. I'll fix it. :+1:

app/src/main/java/chat/rocket/android/util/Extensions.kt
leonardoaramaki: No, we don't need actually. I could rename the property from `content` to `markdown` if you will. We have already a simpler version which is `textContent` but it uses `String` as its type. Actually I don't know what's the real benefit of having this one. I think we could stick just with the framework's `text` property.

app/src/main/java/chat/rocket/android/helper/MessageParser.kt
leonardoaramaki: It's not needed (on this branch). Since `renderMarkdown()` is only called at `MessageViewModel` `init {}` which in turn is only invoked on a suspend function:

https://github.com/RocketChat/Rocket.Chat.Android/blob/b2d415bbadaaca7627802383b20eaa073c45d322/app/src/main/java/chat/rocket/android/chatroom/viewmodel/MessageViewModelMapper.kt#L18-L18

leonardoaramaki: But it's a good point. I'll revisit this later since I think I call `renderMarkdown()` at other places on the other branch.

luciofm: But this suspend function is called from the UI context, so it will block UI anyway...

luciofm: But let's fix it later.

app/src/main/java/chat/rocket/android/util/Extensions.kt
luciofm: I think we should have a simple `content` with `CharSequence` and a `markdown`, just that

app/src/main/java/chat/rocket/android/helper/MessageParser.kt
luciofm: Ok, saw now that it is called from a `CommonPool` context now... But coroutines are so cheap that there is no problem from calling one from another...

leonardoaramaki: No. I fixed the suspended function already. It's using `CommonPool`. See the line above.

leonardoaramaki: @luciofm I set `MessageViewModel` class as a data class but it's everything but not one. I didn't remove the `data` modifier from it solely for the free toString() for debugging. Are you working or are going to work on moving everything to the mapper?

app/src/main/java/chat/rocket/android/helper/MessageParser.kt
```diff
@@ -0,0 +1,176 @@
...
+
+            // pass to super to apply markdown
+            super.visit(blockQuote)
+
+            builder.setSpan(QuoteMarginSpan(context.getDrawable(R.drawable.quote), 10), length, builder.length())
+        }
+    }
+
+    class QuoteMarginSpan(quoteDrawable: Drawable, private var pad: Int) : LeadingMarginSpan, LineHeightSpan {
+        private var drawable: Drawable = quoteDrawable
```
luciofm: Does it need to be mutable? We are changing its attributes, but not the instance itself.

app/src/main/java/chat/rocket/android/dagger/module/AppModule.kt
```diff
@@ -188,12 +188,39 @@ class AppModule {
...
+    @Singleton
+    fun provideConfiguration(context: Application, client: OkHttpClient): SpannableConfiguration {
+        return SpannableConfiguration.builder(context)
+                .asyncDrawableLoader(AsyncDrawableLoader.builder()
+                        .client(client)
+                        .executorService(Executors.newCachedThreadPool())
+                        .resources(context.resources)
+                        .build())
+                .theme(SpannableTheme.builder()
+                        .linkColor(Color.BLUE)
```
filipedelimabrito: Is it possible to use our `colorAccent` here?
https://github.com/RocketChat/Rocket.Chat.Android/blob/d31b9ea1262d750a826e526bb946b7cd8810c797/app/src/main/res/layout/fragment_authentication_log_in.xml#L56

philipbrito: Is it possible to use our `colorAccent` here?
https://github.com/RocketChat/Rocket.Chat.Android/blob/d31b9ea1262d750a826e526bb946b7cd8810c797/app/src/main/res/layout/fragment_authentication_log_in.xml#L56

app/src/main/java/chat/rocket/android/helper/MessageParser.kt
```diff
@@ -0,0 +1,176 @@
...
+            val itop = layout.getLineTop(layout.getLineForOffset(st))
+            val dw = drawable.intrinsicWidth
+            val dh = drawable.intrinsicHeight
+            // XXX What to do about Paint?
+            drawable.setBounds(ix, itop, ix + dw, itop + layout.height)
+            drawable.draw(c)
+        }
+
+        override fun chooseHeight(text: CharSequence, start: Int, end: Int,
+                                  istartv: Int, v: Int,
```
luciofm: what `istartv` means? And I think it has camelCase problems

app/src/main/java/chat/rocket/android/helper/MessageParser.kt
luciofm: @leonardoaramaki yes, the intention is to have the Models just as simple POKOs (Plain Old Kotlin Objects), and move all the work to the mapper and maybe factories if the mapper becomes too complex.

app/src/main/java/chat/rocket/android/dagger/module/AppModule.kt
leonardoaramaki: Sure :+1:

app/src/main/java/chat/rocket/android/helper/MessageParser.kt
leonardoaramaki: Changed

app/src/main/java/chat/rocket/android/helper/MessageParser.kt
leonardoaramaki: It's spanstartv

leonardoaramaki: Those parameters are for measuring. It's `v`for vertical and `spanstartv` on the `LineHeightSpan` interface. I don't think it would look meaningful in camelcase.

rafaelks:
Is this one ready to merge now guys?

PR #733, opened 2018-02-04
[NEW][WIP] Message actions - Reply, Quote, Copy, Delete, Edit and Pin/Unpin

### LINKED ISSUE #721: [NEW] Actions on Message: Copy, Reply, Quote & Pin

leonardoaramaki:
@RocketChat/android

- [x] Reply
- [x] Quote
- [x] Delete
- [x] Copy
- [x] Edit
~- [ ] Share~
- [x] Pin
- [x] Unpin

Also added an PinnedMessagesActivity to show all the pins.

Closes #721 and #718


app/src/main/java/chat/rocket/android/authentication/server/ui/ServerFragment.kt
```diff
@@ -56,6 +57,8 @@ class ServerFragment : Fragment(), ServerView {
         enableUserInput(true)
     }
 
+    override fun showMessage(resId: Int) = showMessage(getString(resId))
```
luciofm: DRY. We should move `showMessage`'s functions to extensions...
Something like
```kotlin
fun Activity.showToast(@StringRes resource: Int, duration: Int = Toast.LENGTH_SHORT) = showToast(getString(resource), duration)
fun Activity.showToast(message: String, duration: Int = Toast.LENGTH_SHORT) = Toast.makeText(this, message, duration).show()

fun Fragment.showToast(@StringRes resource: Int, duration: Int = Toast.LENGTH_SHORT) = showToast(getString(resource), duration)
fun Fragment.showToast(message: String, duration: Int = Toast.LENGTH_SHORT) = activity?.showToast(message, duration)
```

app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomPresenter.kt
```diff
@@ -139,6 +141,122 @@ class ChatRoomPresenter @Inject constructor(private val view: ChatRoomView,
...
+    /**
+     * Delete the message with the given id.
+     *
+     * @param roomId The room id of the message to be deleted.
+     * @param id The id of the message to be deleted.
+     */
+    fun deleteMessage(roomId: String, id: String) {
+        launchUI(strategy) {
+            if (!getPermissionsInteractor.isMessageDeletingAllowed()) {
+                coroutineContext.cancel()
```
luciofm: Do we need to cancel here? I think that just returning is enough.

app/src/main/java/chat/rocket/android/chatroom/ui/BottomSheet.kt
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    /**
+     * Show a menu item and enable it.
+     *
+     * @param itemId The id of the menu item to enable and show.
+     */
+    fun showMenuItem(@IdRes itemId: Int) {
+        val item = menuItems.firstOrNull { it.itemId == itemId }
```
luciofm: A more idiomatic way would be
```kotlin
        menuItems.firstOrNull { it.itemId == itemId }?.apply {
            isVisible = false
            isEnabled = false
        }
```

app/src/main/java/chat/rocket/android/chatroom/ui/BottomSheet.kt
```diff
@@ -0,0 +1,107 @@
...
+        item?.apply {
+            setVisible(true)
+            setEnabled(true)
+        }
+    }
+
+    fun setMenuItemTextColor(@IdRes itemId: Int, @ColorInt color: Int) {
+        val itemIndex = menuItems.indexOfFirst { it.itemId == itemId }
```
luciofm:
```kotlin
        menuItems.firstOrNull() { it.itemId == itemId }?.apply { 
            textColors.put(itemId, color)
        }
```

app/src/main/java/chat/rocket/android/chatroom/ui/BottomSheet.kt
```diff
@@ -0,0 +1,107 @@
...
+    }
+}
+
+class ActionListAdapter(menuItems: List<MenuItem> = emptyList(), callback: MenuItem.OnMenuItemClickListener) :
+        ListBottomSheetAdapter(menuItems = menuItems, callback = callback) {
+
+    override fun onBindViewHolder(holder: ViewHolder, position: Int) {
+        val item = menuItems[position]
+
+        if (showIcons) {
```
luciofm:
```kotlin
if (showIcons) {
            holder.imageIcon.setVisible(item.icon == null)
        } else {
            holder.imageIcon.setVisible(false)
        }
```

app/src/main/java/chat/rocket/android/chatroom/ui/BottomSheet.kt
```diff
@@ -0,0 +1,107 @@
...
+        } else {
+            holder.imageIcon.visibility = View.GONE
+        }
+
+        holder.imageIcon.setImageDrawable(item.icon)
+        holder.textTitle.text = item.title
+        holder.itemView.setOnClickListener {
+            callback?.onMenuItemClick(item)
+        }
+        val color = if (item.itemId == R.id.action_menu_msg_delete) Color.RED else textColors.get(item.itemId)
```
luciofm: We should define these default colors in `colors.xml`

app/src/main/java/chat/rocket/android/chatroom/ui/BottomSheet.kt
```diff
@@ -0,0 +1,107 @@
...
+
+class BottomSheetMenu(adapter: MenuAdapter) : SheetMenu(adapter = adapter) {
+
+    override fun processRecycler(recycler: RecyclerView, dialog: BottomSheetDialog) {
+        if (layoutManager == null) {
+            layoutManager = LinearLayoutManager(recycler.context, LinearLayoutManager.VERTICAL, false)
+        }
+
+        val callback = adapter?.callback
+        adapter?.callback = MenuItem.OnMenuItemClickListener {
```
luciofm: This is really confusing.

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomAdapter.kt
```diff
@@ -50,25 +59,83 @@ class ChatRoomAdapter(private val serverUrl: String) : RecyclerView.Adapter<Chat
         notifyItemInserted(0)
     }
 
-    fun updateItem(index: Int, message: MessageViewModel) {
-        dataSet[index] = message
-        notifyItemChanged(index)
+    fun updateItem(message: MessageViewModel) {
+        val index = dataSet.indexOfFirst { it.id == message.id }
```
luciofm, line 64: We can add an extension here that will be useful on other places.
```kotlin
public inline fun <T> List<T>.firstWithIndex(predicate: (T) -> Boolean): Pair<Int, T>? {
    val index = this.indexOfFirst(predicate)
    val data = if (index != -1) this[index] else null

    return if (data != null) Pair(index, data) else null
}

    fun updateItem(message: MessageViewModel) {
        dataSet.firstWithIndex { it.id == message.id }?.let { (index, _) ->
            dataSet[index] = message
            notifyItemChanged(index)
        }
    }
```

app/src/main/java/chat/rocket/android/chatroom/ui/PinnedMessagesAdapter.kt
```diff
@@ -0,0 +1,144 @@
...
+                    }
+                }
+
+                image_attachment.setVisible(imageVisible)
+                audio_video_attachment.setVisible(videoVisible)
+                file_name.text = message.attachmentTitle
+            }
+        }
+
+        private fun bindUserAvatar(message: MessageViewModel, drawee: SimpleDraweeView, imageUnknownAvatar: ImageView) = message.getAvatarUrl(serverUrl).let {
```
luciofm:
```kotlin
fun function() = statement()
```
works better for one liners... If it has multiple lines, lets use `{ }` and `return`

app/src/main/java/chat/rocket/android/server/domain/GetChatRoomsInteractor.kt
```diff
@@ -19,4 +20,21 @@ class GetChatRoomsInteractor @Inject constructor(private val repository: ChatRoo
...
+    /**
+     * Get a specific room by its id.
+     *
+     * @param serverUrl The server url where the room is.
+     * @param roomId The id of the room to get.
+     *
+     * @return The ChatRoom object or null if we couldn't find any.
+     */
+    suspend fun getById(serverUrl: String, roomId: String): ChatRoom? {
+        return async(CommonPool) {
```
luciofm: return `withContext(CommonPool) { }`  is the same as `return async(CommonPool) { }.await()`

luciofm (changes requested):
Added some comments

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomAdapter.kt
leonardoaramaki: I see what you mean but I think it makes it less clear.

app/src/main/java/chat/rocket/android/chatroom/ui/BottomSheet.kt
leonardoaramaki: The `adapter` property is inherited from the superclass `SheetMenu` from the library with the same name and all its properties, including the `callback` are nullable (library is written in Kotlin). 
It's possible although unlikely that a user of this class will drop up a list of items with no callback. At least the author of the library permitted this. 
I did some changes though. I'm pushing right away.

app/src/main/java/chat/rocket/android/authentication/server/ui/ServerFragment.kt
leonardoaramaki: But on all those fragments the `showMessage` method is only accessible by the view. So we would still repeat ourselves but on a different manner:

Instead of 
```kotlin
    override fun showMessage(resId: Int) = showMessage(getString(resId))

    override fun showMessage(message: String) = Toast.makeText(activity, message, Toast.LENGTH_SHORT).show()
```
we would have
```kotlin
    override fun showMessage(resId: Int) = showToast(resId)
    
    override fun showMessage(message: String) = showToast(message)
```
On the other hand, if by DRY you mean repeating the `Toast.makeText(...).show()` than I think it's valid. :+1:

luciofm (approved):

PR #734, opened 2018-02-06
[NEW] Show groupable messages.

philipbrito:
Note: Groupable attribute for messages should be returned by the SDK. Currently all messages are set as `groupable=false`

### DIFF
--- a/app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomAdapter.kt
+++ b/app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomAdapter.kt
@@ -3,7 +3,6 @@ package chat.rocket.android.chatroom.ui
 import android.support.v7.widget.RecyclerView
 import android.view.View
 import android.view.ViewGroup
-import android.widget.ImageView
 import android.widget.TextView
 import chat.rocket.android.R
 import chat.rocket.android.chatroom.viewmodel.AttachmentType
@@ -11,33 +10,33 @@ import chat.rocket.android.chatroom.viewmodel.MessageViewModel
 import chat.rocket.android.player.PlayerActivity
 import chat.rocket.android.util.inflate
 import chat.rocket.android.util.setVisible
-import chat.rocket.common.util.ifNull
 import com.facebook.drawee.view.SimpleDraweeView
 import com.stfalcon.frescoimageviewer.ImageViewer
 import kotlinx.android.synthetic.main.avatar.view.*
 import kotlinx.android.synthetic.main.item_message.view.*
 import kotlinx.android.synthetic.main.message_attachment.view.*
 
-class ChatRoomAdapter(private val serverUrl: String) : RecyclerView.Adapter<ChatRoomAdapter.ViewHolder>() {
+class ChatRoomAdapter: RecyclerView.Adapter<ChatRoomAdapter.ViewHolder>() {
+    private val dataSet = ArrayList<MessageViewModel>()
 
     init {
         setHasStableIds(true)
     }
 
-    val dataSet = ArrayList<MessageViewModel>()
-
     override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): ViewHolder =
-            ViewHolder(parent.inflate(R.layout.item_message), serverUrl)
+            ViewHolder(parent.inflate(R.layout.item_message))
 
-    override fun onBindViewHolder(holder: ViewHolder, position: Int) = holder.bind(dataSet[position])
+    override fun onBindViewHolder(holder: ViewHolder, position: Int) =
+        holder.bind(dataSet[position], dataSet.getOrNull(position + 1))
 
-    override fun onBindViewHolder(holder: ViewHolder, position: Int, payloads: MutableList<Any>?) {
-        onBindViewHolder(holder, position)
-    }
+    override fun getItemCount(): Int =
+        dataSet.size
 
-    override fun getItemCount(): Int = dataSet.size
+    override fun getItemViewType(position: Int): Int =
+        position
 
-    override fun getItemViewType(position: Int): Int = position
+    override fun getItemId(position: Int): Long =
+        dataSet[position].id.hashCode().toLong()
 
     fun addDataSet(dataSet: List<MessageViewModel>) {
         val previousDataSetSize = this.dataSet.size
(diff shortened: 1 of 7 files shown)


rafaelks:
@filipedelimabrito I believe the API is sending groupable=false always. On iOS, we are grouping all messages from the same user on the interval of 15 minutes. If there's an attachment, message is not groupable. If there's a system message, message is not groupable.

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomAdapter.kt
```diff
@@ -3,41 +3,40 @@ package chat.rocket.android.chatroom.ui
...
+        holder.bind(dataSet[position], dataSet.getOrNull(position + 1))
 
-    override fun onBindViewHolder(holder: ViewHolder, position: Int, payloads: MutableList<Any>?) {
-        onBindViewHolder(holder, position)
-    }
+    override fun getItemCount(): Int =
+        dataSet.size
 
-    override fun getItemCount(): Int = dataSet.size
+    override fun getItemViewType(position: Int): Int =
```
luciofm, line 35: That way you will never recycle a View on the RecyclerView.

https://developer.android.com/reference/android/support/v7/widget/RecyclerView.Adapter.html#getItemViewType(int)

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomAdapter.kt
```diff
@@ -55,22 +54,28 @@ class ChatRoomAdapter(private val serverUrl: String) : RecyclerView.Adapter<Chat
...
-            text_user_name.text = message.sender
+        fun bind(message: MessageViewModel, nextMessage: MessageViewModel?) = with(itemView) {
+            image_avatar.setImageURI(message.avatarUri)
+            text_sender.text = message.senderName
             text_message_time.text = message.time
-            text_content.text = message.content
 
-            bindAttachment(message, message_attachment, image_attachment, audio_video_attachment,
-                    file_name)
+            if (nextMessage != null) {
```
luciofm, line 64: I think we should group messages on the ViewModel/ViewModel Mapper and keep the Adapter logic free. What do you thing @leonardoaramaki

app/src/main/java/chat/rocket/android/chatroom/viewmodel/MessageViewModel.kt
```diff
@@ -56,36 +60,55 @@ data class MessageViewModel(val context: Context,
...
-            is UserLeft -> contentMessage = getSystemMessage(context.getString(R.string.message_user_left))
-            is UserAdded -> contentMessage = getSystemMessage(
-                    context.getString(R.string.message_user_added_by, message.message, message.sender?.username))
-            is RoomNameChanged -> contentMessage = getSystemMessage(
-                    context.getString(R.string.message_room_name_changed, message.message, message.sender?.username))
-            is UserRemoved -> contentMessage = getSystemMessage(
-                    context.getString(R.string.message_user_removed_by, message.message, message.sender?.username))
-            else -> contentMessage = getNormalMessage()
+            //TODO: Add implementation for Welcome type.
+            is MessageRemoved -> {
```
luciofm, line 85: For one line statements it is ok to go inline and leave the `{}` out (https://kotlinlang.org/docs/reference/coding-conventions.html#method-call-formatting)

app/src/main/java/chat/rocket/android/chatroom/viewmodel/MessageViewModel.kt
```diff
@@ -94,10 +117,14 @@ data class MessageViewModel(val context: Context,
 
     private fun getSystemMessage(content: String): CharSequence {
         val spannableMsg = SpannableString(content)
-        spannableMsg.setSpan(StyleSpan(Typeface.ITALIC), 0, spannableMsg.length,
-                0)
-        spannableMsg.setSpan(ForegroundColorSpan(Color.GRAY), 0, spannableMsg.length,
-                0)
+        spannableMsg.setSpan(StyleSpan(Typeface.ITALIC),
```
luciofm, line 120: both can be on a single line, they fit in 100 columns.

luciofm: And when needed, you can group related parameters on the same line (https://kotlinlang.org/docs/reference/coding-conventions.html#method-call-formatting)

luciofm (changes requested):

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomAdapter.kt
filipedelimabrito: ? both `getItemCount()` and `getItemViewType()` is here. 
What are you referring to?

app/src/main/java/chat/rocket/android/chatroom/viewmodel/MessageViewModel.kt
philipbrito: Just make this change because it is more readable. Do you agree?

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomAdapter.kt
filipedelimabrito: ? both `getItemCount()` and `getItemViewType()` is here. 
What are you referring to?

app/src/main/java/chat/rocket/android/chatroom/viewmodel/MessageViewModel.kt
luciofm: I actually do prefer the one line version

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomAdapter.kt
luciofm:
```kotlin
    override fun getItemViewType(position: Int): Int =
        position
```
You are using the position for the View Type, each row will have a different view type, preventing the RecyclerView to recycler any view.

luciofm: And we are actually not using different view types here.

luciofm:
@leonardoaramaki @filipedelimabrito I vote to close this one and implement this on the Mapper.

I'm currently creating a BaseViewModel and several ViewModels for the messages list. One for the message itself, one for image attachments, one for video, for audio, for Url preview, for message quoting (response), so they will all have different view types. We can create a new GroupedMessageViewModel, that has just the text content, no thumb and timestamp..

luciofm:
Will be handled on the viewmodel mapper

PR #741, opened 2018-02-09
[NEW] File upload

philipbrito:


filipedelimabrito:
@rafaelks Does the Rocket.Chat REST API support the upload for any kind of file or only for image, audio and video?

From [here](https://rocket.chat/docs/developer-guides/rest-api/chat/postmessage/#attachment-field-objects), the right answer is only for image, audio and video?

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
```diff
@@ -37,15 +41,21 @@ private const val BUNDLE_CHAT_ROOM_ID = "chat_room_id"
...
     private lateinit var chatRoomName: String
     private lateinit var chatRoomType: String
     private var isChatRoomReadOnly: Boolean = false
     private lateinit var adapter: ChatRoomAdapter
+    // For reveal and unreveal anim.
+    private val hypotenuse by lazy { Math.hypot(relative_layout.width.toDouble(), relative_layout.height.toDouble()).toFloat() }
+    private val max by lazy { Math.max(layout_message_attachment_options.width.toDouble(), layout_message_attachment_options.height.toDouble()).toFloat() }
+    private val centerX by lazy { recycler_view.right }
+    private val centerY by lazy { recycler_view.bottom }
+    private lateinit var handler: Handler
```
luciofm: You can initialize the handler here. As a immutable `val`

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
```diff
@@ -37,15 +41,21 @@ private const val BUNDLE_CHAT_ROOM_ID = "chat_room_id"
...
 class ChatRoomFragment : Fragment(), ChatRoomView {
-
     @Inject lateinit var presenter: ChatRoomPresenter
     private lateinit var chatRoomId: String
     private lateinit var chatRoomName: String
     private lateinit var chatRoomType: String
     private var isChatRoomReadOnly: Boolean = false
     private lateinit var adapter: ChatRoomAdapter
+    // For reveal and unreveal anim.
+    private val hypotenuse by lazy { Math.hypot(relative_layout.width.toDouble(), relative_layout.height.toDouble()).toFloat() }
```
luciofm: Maybe use a custom view and have an `attach` method passing the anchor view. Also, `relative_layout` is not very descriptive

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
```diff
@@ -102,19 +121,40 @@ class ChatRoomFragment : Fragment(), ChatRoomView {
         }
     }
 
+    override fun uploadFile(uri: Uri) {
+        var fileName: String? = null
+        var mimeType: String? = null
+        var fileRealPath: String? = null
+        activity?.apply {
+            fileName = uri.getFileName(this)
```
luciofm: yout can move the declaration down here, and use immutable `val`
```kotlin
val filename = uri.getFileName(this)
```

Maybe move all this to the presenter, just call `presenter.uploadFile(uri, message)` and create an interactor for retrieving the information from the `Uri`

It is better for handling errors, and calling `view.showInvalidFile()` for example...

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
```diff
@@ -134,9 +174,68 @@ class ChatRoomFragment : Fragment(), ChatRoomView {
...
+                    hideAttachmentOptions()
+                } else {
+                    showAttachmentOptions()
+                }
+            }
+
+            view_dim.setOnClickListener { hideAttachmentOptions() }
+
+            button_files.setOnClickListener {
+                handler.postDelayed({
```
luciofm, line 265: `handler.postDelayed()` with an anonymous `Runnable` is a source of a lot of crashes.
If you rotate the screen, or finish the activity before the delay, it will crashe accessing an invalid Activity and invalid views.

luciofm (changes requested):

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
philipbrito: Why not after the fragment (on `onViewCreated()`) is created?

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
philipbrito: I did human tests here and get no crash when rotating the screen, finishing the activity, etc. Could you check it too?

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
luciofm: Because it doesn't need to wait until `onViewCreated` and it can be immutable...

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
luciofm: It is a race condition and it will crash at some point...

When using `postDelayed` you should always call `removeCallbacks(runnable)` on the appropriate lifecycle callback.

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
philipbrito: Because we are using the handler after `onViewCreated()` is called we don't need to keep that variable member in the memory when the `Fragment` is created.

Maybe initialising the handler as proposed by you has a minimum impact, but I believe that we don't need to create it if we are using only after the UI is created...

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
philipbrito: Thank you @luciofm 👍

app/src/main/java/chat/rocket/android/authentication/signup/presentation/SignupPresenter.kt
```diff
@@ -50,8 +50,16 @@ class SignupPresenter @Inject constructor(private val view: SignupView,
                         view.showLoading()
 
                         try {
-                            client.signup(email, name, username, password) // TODO This function returns a user so should we save it?
-                            client.login(username, password) // TODO This function returns a user token so should we save it?
+                            client.signup(
```
luciofm: Both functions fit within 100 columns. If the comment is the problem, we can move it up to a single line
```kotlin
// TODO This function returns a user so should we save it?
client.signup(email, name, username, password)
```

app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomPresenter.kt
```diff
@@ -82,13 +97,39 @@ class ChatRoomPresenter @Inject constructor(private val view: ChatRoomView,
                     view.showMessage(it)
                 }.ifNull {
                     view.showGenericErrorMessage()
-                }
+                    view.showGenericErrorMessage()
```
luciofm: Seems to be a merge error, calling 2 times `view.showGenericErrorMessage()` and bad indentation on the closing `}`

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
luciofm: It won't have any noticeable impact, one because it will be created when we need the fragment, doesn't matter if on `onViewCreated` or on the Object creation, and because it is a simple object instantiation...

Ideally we should use immutables and readonly `val`s whenever it is possible, since it prevents all sort of bugs down the road.

>Classes should be immutable unless there’s a very good reason to make them mutable.
Joshua Block, Effective Java

app/src/main/java/chat/rocket/android/authentication/login/presentation/LoginPresenter.kt
```diff
@@ -108,8 +111,8 @@ class LoginPresenter @Inject constructor(private val view: LoginView,
                                     exception.message?.let {
                                         view.showMessage(it)
                                     }.ifNull {
-                                        view.showGenericErrorMessage()
-                                    }
+                                            view.showGenericErrorMessage()
```
luciofm: There are a lot of indentation problems like this, on `ifNull { }`

app/src/main/java/chat/rocket/android/authentication/signup/presentation/SignupPresenter.kt
```diff
@@ -60,8 +68,8 @@ class SignupPresenter @Inject constructor(private val view: SignupView,
                             exception.message?.let {
                                 view.showMessage(it)
                             }.ifNull {
-                                view.showGenericErrorMessage()
-                            }
+                                    view.showGenericErrorMessage()
```
luciofm: `ifNull` indentation

app/src/main/java/chat/rocket/android/authentication/twofactor/presentation/TwoFAPresenter.kt
```diff
@@ -54,8 +58,8 @@ class TwoFAPresenter @Inject constructor(private val view: TwoFAView,
                                 exception.message?.let {
                                     view.showMessage(it)
                                 }.ifNull {
-                                    view.showGenericErrorMessage()
-                                }
+                                        view.showGenericErrorMessage()
```
luciofm: `ifNull` indentation

app/src/main/java/chat/rocket/android/chatroom/presentation/PinnedMessagesPresenter.kt
```diff
@@ -38,14 +38,16 @@ class PinnedMessagesPresenter @Inject constructor(private val view: PinnedMessag
...
                     pinnedMessagesListOffset = pinnedMessages.offset.toInt()
-                    val messageList = mapper.mapToViewModelList(pinnedMessages.result, settings).filterNot { it.isSystemMessage }
+                    val messageList = mapper.mapToViewModelList(pinnedMessages.result, settings)
+                        .filterNot { it.isSystemMessage }
                     view.showPinnedMessages(messageList)
                     view.hideLoading()
                 }.ifNull {
-                    Timber.e("Couldn't find a room with id: $roomId at current server.")
-                }
+                        Timber.e("Couldn't find a room with id: $roomId at current server.")
```
luciofm: `ifNull` indentation

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomAdapter.kt
```diff
@@ -29,26 +27,24 @@ import ru.whalemare.sheetmenu.extension.toList
 class ChatRoomAdapter(private val roomType: String,
                       private val roomName: String,
                       private val presenter: ChatRoomPresenter) : RecyclerView.Adapter<ChatRoomAdapter.ViewHolder>() {
+    private val dataSet = ArrayList<MessageViewModel>()
```
luciofm, line 30: Please revert https://github.com/RocketChat/Rocket.Chat.Android/pull/741/commits/05d9453c7ffbce2e352b72b058b33d832e986b7f, grouped messages are being handled on the Mapper + ViewModels on a future PR (awaiting to merge this one, to rebase and send it)

luciofm (changes requested):
Please revert https://github.com/RocketChat/Rocket.Chat.Android/pull/741/commits/05d9453c7ffbce2e352b72b058b33d832e986b7f
Grouped messages will be handled with ViewModels (and a special Mapping). besides the ChatRoomAdapter is currently broken on 05d9453 as it doesn't recycle any view.

app/src/main/java/chat/rocket/android/authentication/login/presentation/LoginPresenter.kt
philipbrito: It's is crazy...

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomAdapter.kt
philipbrito: Sure!

konishi-t:
File Upload Error:
Network Error: Write error: ssl=0x7f67d07e00:
I/O error during system call, Broken pipe
known issue?

Android(git clone 2018/02/21) --(https)-> nginx --(http)-> Rocket.Chat 0.61.0

PR #776, opened 2018-02-17
[NEW][FIX] Prevents losing all entered text when a new message arrives on the chat

### LINKED ISSUE #757: [BUG] When typing a message and someone sends a message, it clears the text being typed

philipbrito:
Also,

- Keeps the keyboard open after sending a message.
- Shows the entire new message when it arrives on the chat room.
- Adds a widget to easily show the latest messages when moving the list up.

Closes #757

### DIFF
--- a/app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomView.kt
+++ b/app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomView.kt
@@ -75,7 +75,19 @@ interface ChatRoomView : LoadingView, MessageView {
      */
     fun showEditingAction(roomId: String, messageId: String, text: String)
 
-    fun disableMessageInput()
+    /**
+     * Disabling the send message button avoids the user tap this button multiple
+     * times to send a same message.
+     */
+    fun disableSendMessageButton()
 
-    fun enableMessageInput(clear: Boolean = false)
+    /**
+     * Enables the send message button.
+     */
+    fun enableSendMessageButton()
+
+    /**
+     * Clears the message composition.
+     */
+    fun clearMessageComposition()
 }
--- a/app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
+++ b/app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
@@ -24,6 +24,8 @@ import dagger.android.support.AndroidSupportInjection
 import kotlinx.android.synthetic.main.fragment_chat_room.*
 import kotlinx.android.synthetic.main.message_attachment_options.*
 import kotlinx.android.synthetic.main.message_composer.*
+import kotlinx.android.synthetic.main.message_list.*
+import timber.log.Timber
 import javax.inject.Inject
 
 fun newInstance(chatRoomId: String, chatRoomName: String, chatRoomType: String, isChatRoomReadOnly: Boolean): Fragment {
@@ -85,7 +87,10 @@ class ChatRoomFragment : Fragment(), ChatRoomView {
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
         presenter.loadMessages(chatRoomId, chatRoomType)
-        setupComposer()
+
+        setupRecyclerView()
+        setupFab()
+        setupMessageComposer()
         setupActionSnackbar()
     }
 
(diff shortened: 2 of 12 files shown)


luciofm:
@filipedelimabrito nice one :smiley:, will take a look later for merging

app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomView.kt
```diff
@@ -80,9 +80,21 @@ interface ChatRoomView : LoadingView, MessageView {
      */
     fun showEditingAction(roomId: String, messageId: String, text: String)
 
-    fun disableMessageInput()
+    /**
+     * Disabling the send message button avoids the user tap this button multiple
+     * times to send a same message.
+     */
+    fun disableSendMessageButton()
```
luciofm, line 87: I think we should disable the input also... It is weird to edit the message while it is being sent...
Maybe we can use an overlay on top of the `EditText` so we don't need to disable it and get the keyboard closed

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
```diff
@@ -227,35 +235,44 @@ class ChatRoomFragment : Fragment(), ChatRoomView {
         showMessage(getString(R.string.max_file_size_exceeded, fileSize, maxFileSize))
     }
 
-    private fun setupComposer() {
+    private fun setupRecyclerView() {
+        recycler_view.addOnScrollListener(object : RecyclerView.OnScrollListener() {
+            override fun onScrolled(recyclerView: RecyclerView, dx: Int, dy: Int) {
+                super.onScrolled(recyclerView, dx, dy)
```
luciofm: There is no need to call `super` here

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
```diff
@@ -281,15 +300,26 @@ class ChatRoomFragment : Fragment(), ChatRoomView {
...
     }
 
-    private fun clearActionMessage() {
-        citation = null
-        editingMessageId = null
-        text_message.text.clear()
-        actionSnackbar.dismiss()
+    private fun subscribeTextMessage() {
+        var playAnimation = true
+        text_message.asObservable(0)
```
luciofm: We need to dispose this subscription when destroying this Fragment's view.

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
```diff
@@ -281,15 +300,26 @@ class ChatRoomFragment : Fragment(), ChatRoomView {
...
+            .subscribe({ t ->
```
luciofm: Create a function with a meaningful name, something like

`text_message.asObservable(0).subscribe { setupComposeButtons(it) }`

It increases the readability of the code

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
```diff
@@ -281,15 +300,26 @@ class ChatRoomFragment : Fragment(), ChatRoomView {
...
+                if (t.isNotEmpty() && playAnimation) {
+                    button_show_attachment_options.fadeInOrOut(1F, 0F, 120)
```
luciofm: I think it is better to have  `fadeIn` and `fadeOut` functions, it increases the readability of the code, and takes just one parameter (or none, if using the default duration)

app/src/main/java/chat/rocket/android/chatrooms/ui/ChatRoomsAdapter.kt
```diff
@@ -18,6 +18,7 @@ import chat.rocket.core.model.ChatRoom
 import com.facebook.drawee.view.SimpleDraweeView
 import kotlinx.android.synthetic.main.avatar.view.*
 import kotlinx.android.synthetic.main.item_chat.view.*
+import kotlinx.android.synthetic.main.unread_messages_badge.view.*
```
luciofm, line 18: Do we need this include? don't see any changes on the adapter using it

app/src/main/res/drawable/ic_arrow_downward_black_24dp.xml
```diff
@@ -0,0 +1,11 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportHeight="24.0"
+    android:viewportWidth="24.0">
+
+    <path
+        android:fillColor="#010101"
```
luciofm: Maybe remove the black from the file name, since it is not actually black

luciofm (changes requested):

app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomView.kt
philipbrito: I think that we always should see how the other app does the things. Take a look on Telegram, it never closes the keyboard when sending, editing or replaying a message (during and after).

luciofm: I'm not saying to close the keyboard. I was talking to not let the user edit the message while we are sending it.

Maybe we could attach a `TextWatcher` that prevents edits while the message is being sent.

app/src/main/res/drawable/ic_arrow_downward_black_24dp.xml
philipbrito: Good catch :mag: 
👍

philipbrito: Interesting @luciofm 
This drawable comes from the official material icon pack. They are naming the file as `...black_24dp` but as pointed by you that file isn't black. Could you report to the Android team too? :grimacing:

Thanks for pointing.

app/src/main/java/chat/rocket/android/chatrooms/ui/ChatRoomsAdapter.kt
philipbrito: We need because `text_total_unread_messages` comes from the `unread_messages_badge` layout.

Note: we'll use the badge on "chat list" view and "chat room" view.

PR #786, opened 2018-02-21
[WIP][RFC] Remove SVG support

luciofm:
[Request for Comments]

Remove SVG support in favor of a simple and efficient `AvatarTextDrawable`

Also use Glide instead of of Fresco:
- It doesn't needs a custom view type, it works with regular `ImageView`s or any target you can think of. This is good when working with Activity Transitions (going from the attachment in the room to the fullscreen viewer with a nice transition).
- It also supports loading images from different sources, like the video preview PR, we end up needing an extra `ImageView` for the preview, when it could be the same as for images attachment.
- It was really easy to write the decoder and transcoder
- Its fluent API is a big win :)
- Also reduced the dex method count by 3k

And works with 0.61 also


luciofm:
Also, from my tests (need to check it better), it uses considerably less memory with glide+drawable then fresco+svg

philipbrito:
@luciofm I think that you could also delete the `ForFresco.kt` class.

luciofm:
opening a new one

PR #787, opened 2018-02-22
[NEW] Password Update Feature - Issue #755

### LINKED ISSUE #755: [NEW] Update user password
From the setting menu (yeah, we need to add this to our navigation drawer), allow to the user update its password through [client.updateProfile(...)](https://github.com/RocketChat/Rocket.Chat.Kotlin.SDK/blob/99b4ce6c2e79a7c6f7a4d2a9e3a8b2b719b94720/core/src/main/kotlin/chat/rocket/core/internal/rest/User.kt#L47).

Tip: see the `updateProfile()` in action [here](https://github.com/RocketChat/Rocket.Chat.Android/blob/befa16fcbb18a5c862222ff0303c2ea2f6cc7f62/app/src/main/java/chat/rocket/android/profile/presentation/ProfilePresenter.kt#L52).

samrmur:
@RocketChat/android

Closes #755 

**Update Information**

I have completed the main functionality for the password update feature. I have done my best to keep it consistent with the project format and the programming style. In terms of what I have done, I have added a **settings option** in the navigation drawer with the only current option being _Change Password_ 
 
The **password screen** itself currently only contains two text fields for changing the user's password, one to enter the new password and one to confirm it so that the user does not mess up their password. The update button is available in the Action Mode, which _currently_ only activates when both text fields contain content with a length larger than 5 and both text fields have the same content.
 

**Possible New Feature**

- Adding password limitations so that the user is forced to enter a strong password (must have uppercase letters, a symbol, ...)

**Extra**

- Sorry this took a while, had schoolwork and spent an entire day trying to solve Dex errors (which ended up getting fixed later on). 

- I also understand that the application has a Portuguese language set. I am unfamiliar with the language and, if possible, would like to be provided with the translations of the following _strings_ or a place where I can get the proper translations from so that I can add them to the list.

  - Enter New Password

  - Confirm New Password

  - Confirm Password Change

  - Settings

  - Change Password


philipbrito:
Looking very good @samrmur  🎉 
We'll review it later.

About the strings to be translated to Portuguese:

- Enter New Password -> Informe a nova senha
- Confirm New Password -> Confirme a nova senha
- Confirm Password Change -> Confirme a alteração da senha
- Settings -> Configurações
- Change Password -> Alterar senha

Thank you!

rafaelks:
@filipedelimabrito Is this one ready to merged?

filipedelimabrito:
Thanks for this PR @samrmur !

app/src/main/java/chat/rocket/android/settings/presentation/SettingsView.kt
```diff
@@ -0,0 +1,3 @@
+package chat.rocket.android.settings.presentation
+
+interface SettingsView
```
philipbrito, line 3: Why adding it? This interface contains no function, so it is useless.

philipbrito:
Thanks for this PR @samrmur !

app/src/main/java/chat/rocket/android/settings/presentation/SettingsView.kt
samrmur: Added the interface to stick to project structure and if the interface itself needs to have specific functions in the future. Also added in case the implemented view goes through a major change, meaning all files referencing the view have to be changed as well (this allows for less work).

PR #804, opened 2018-02-23
[NEW] Mark room as read.

philipbrito:
Depends on: https://github.com/RocketChat/Rocket.Chat.Kotlin.SDK/pull/73


rafaelks (approved):

geekgonecrazy:
Would love to see this make it in with #796 would actually be usable as daily driver.  Still have to go open in old app to mark channels as read

PR #809, opened 2018-02-24
[NEW] feat: Add Swipe to refresh in ChatRooms

### LINKED ISSUE #808: [NEW] Add Swipe to refresh in ChatRooms
Your Rocket.Chat.Android version: 2.0.0dev6

**Actual behaviour**
Currently there is no option to refresh ChatRooms. Suppose user has slow internet connection then sometimes it becomes hectic to restarting app or switching between fragment.

**Expected behaviour**
Add feature to refresh ChatRooms on swipe

**Willing to work on this issue?**
Yes

Shailesh351:
@RocketChat/android

Closes #808 

**Changes**
- Added SwipeRefreshLayout as parent of RecyclerView

**Preview**

video2gif_20180224_094315


geekgonecrazy:
Even if this did nothing.  I love it!  Something about pull to refresh just makes a person feel happy and that they aren't missing something 😀

app/src/main/java/chat/rocket/android/chatrooms/ui/ChatRoomsFragment.kt
```diff
@@ -51,6 +51,10 @@ class ChatRoomsFragment : Fragment(), ChatRoomsView {
         setupToolbar()
         setupRecyclerView()
         presenter.loadChatRooms()
+        swipe_refresh_layout.setOnRefreshListener {
+            presenter.loadChatRooms()
+            swipe_refresh_layout.isRefreshing = false
```
luciofm: You should set `swipe_refresh_layout.isRefreshing = false` at `updateChatRooms()`

luciofm: Also we should or prevent the Swito to Refresh when using the search, or reapply the search query or clear the search.

luciofm (changes requested):

app/src/main/java/chat/rocket/android/chatrooms/ui/ChatRoomsFragment.kt
Shailesh351: I thinked of setting `swipe_refresh_layout.isRefreshing = false` at `updateChatRooms()` but problem with that is we already have three-dot loading indicator, so it won't look good showing two loading indicator.

Will update PR soon for clearing search.

rafaelks:
@luciofm This one is ready to be reviewed again I think.

rafaelks:
This build is crashing for me after the connect screen. Can you take a look please @Shailesh351?

Shailesh351:
@rafaelks  Rebased PR on develop-2.x and also fixed grouping while searching have a look on this.
Thanks

rafaelks:
@Shailesh351 I'll close this PR, since we don't plan to add this feature for now in the app. Thank you very much! 👍

PR #815, opened 2018-02-24
[IMPROVEMENT] Use glide, remove fresco and SVG

luciofm:


rafaelks:
@luciofm Lots of conflicts. 🙈

philipbrito:
@luciofm I think that this PR could be closed since the SVG support was already removed and the plan is to use the Picasso3 lib instead glide.
Am I right?

rafaelks:
@luciofm Closing this one for now, looks like it's inactive. If we need to open again, let's do it.

PR #823, opened 2018-02-28
[IMPROVEMENT] Fixed Layout for Login Screen

### LINKED ISSUE #819: [IMPROVEMENT] Social accounts interface is shown even when it should be hidden
- When you enter a wrong username/password on the login screen the social accounts interface is shown (even when it should be hidden) -> we should fix it 

- After entering a valid username and password the screen goes to the TwoFA or Chat List screen but the social interface that was hidden is shown at this time -> we should fix it.

samrmur:
@RocketChat/android

Closes #819 

**Information**
This pull request fixes the issue of having the social media interface show up after a failed log-in or just as the fragment is switching. It also fixes the same problem for when a user goes back from the TwoFA screen to the Login screen.

**Changes Made**

- Removed checking whether the social interface needed to be displayed or not from the `enableUserInput(value: Boolean)` method (this is not needed since the login button isn't displayed until the user has either entered text in the user-name or password or is focused on a text field and when there is no text, an API call isn't even attempted).
- Created `areLoginOptionsNeeded()` to check if the login button, social media interface, or sign up view are needed. Used this method for the listener and when the view state is retrieved.

**Screenshot**

### DIFF
--- a/app/src/main/java/chat/rocket/android/authentication/login/ui/LoginFragment.kt
+++ b/app/src/main/java/chat/rocket/android/authentication/login/ui/LoginFragment.kt
@@ -12,6 +12,7 @@ import android.view.ViewGroup
 import android.view.ViewTreeObserver
 import android.widget.ImageButton
 import android.widget.ScrollView
+
 import chat.rocket.android.R
 import chat.rocket.android.authentication.login.presentation.LoginPresenter
 import chat.rocket.android.authentication.login.presentation.LoginView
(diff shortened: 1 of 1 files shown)


rafaelks:
@samrmur Thank you for the fix! The CI is not passing, can you take a look? We'll review your code soon.

samrmur:
@rafaelks Looks the build just failed because it reached memory limit, re-run worked properly.

philipbrito:
Thanks for this PR @samrmur !
We'll review it later!
Thank you!

philipbrito (approved):

PR #871, opened 2018-03-09
[NEW] CAS login support.

philipbrito:


rafaelks (changes requested):
There are some things to be fixed here...

- The app doesn't authenticate the user; It's crashing the app after authenticating on the webview;
- The user isn't prompted to actually open CAS, you're opening directly. See how it should work by coming here: https://chat.hyperledger.org/home. It shows a button and let the user choose, because sometimes the server can support CAS and other options;
- What do you think about using a fullscreen web view instead of just part of the screen?

philipbrito:
@rafaelks Is there more than one button on the login screen for servers that authenticates with CAS? I mean, is there the possibility to the server show more than one "CAS login URL"?. We are getting only one URL from the server settings so I think that we don't need to make a button to let the user open the webview (it is a pain).

Opening a fullscreen webview can be great. What do you think about after connecting with the server, we show the fullscreen webview (without the needed to click in a button to show it)?. It works for you?

P.s: I guess - from the code - that the web app shows a new window to the user log in the CAS because it is "listening to" the close window action (that is trigged after the user successful authentic in the CAS) - we don't need to add this behavior to our *mobile* app.

rafaelks:
@filipedelimabrito Server can have multiple authentication methods, if we open the web view directly, we won't support all of them correctly. So we need to have the buttons for now and can think about not showing the buttons if we check that they've only one authentication method.

Regarding the fullscreen: having a fullscreen experience on the web view can be much better because some websites may not have a good experience in mobile apps. 

Let's change the experience of this screen to:

1. Add the button for now;
2. When user tap into the button, open the webview in fullscreen with an "X" button to cancel the operation;
3. When we get the token, you close the webview window and shows a loader inside of the CAS login button;

What do you think?

philipbrito:
I did the improvements in the login view (as well as adding one more authentication server settings). Check if it works for you (the UI and if it is crash-free now).

PR #889, opened 2018-03-12
[ENHANCEMENT] Scroll to message when keyboard opened

### LINKED ISSUE #872: [ENHANCEMENT] Messages goes hidden when trigger keyboard for new message
Your Rocket.Chat.Android version: **2.0.0-dev9**

Mobile device model and OS version: Samsung S8

Noticed this, when typing new message/reply in channel/dm the keyboard hide the chat view. It suppose should push the chat view up to see the latest messages.

shriakhilc:
@RocketChat/android

Closes #872 

The existing `onKeyboardOpened` callback had an issue where the first click was not detected by `performClick()`. Further, eventual clicks on the text field while the keyboard were open also triggered the event.

A boolean was added to ensure the callback only triggered when the keyboard was not already open. To handle the first click, `onFocusChanged` was overridden and `performClick` called through it manually.

In `ChatRoomFragment.kt`, the `scrollToPosition` method was not behaving as intended due to being executed before the resizing of views. Hence, the method was put into a `Handler` and posted after a delay.  (A similar method call exists in the `showReplyingAction()` function [here](https://github.com/TheGamer007/Rocket.Chat.Android/blob/92cc58c3b92b8eb146bd10eb65d4debb4be505d9/app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt#L262), but it was left unchanged since no issues were reported with it).

Based on the discussion at the original issue, a message threshold of `3` has been applied below which it scrolls directly to the bottom. Above this, it scrolls to the bottom-most visible element. This can be tweaked a bit to center a more appropriate item, like the one above it or the bottom-most _completely_ visible element. 

The only potential issue I've seen with scrolling for each item is that if a message has very long content, then it tends to scroll either to the top or bottom of that message. For normal messages, it neatly brings them into view.

### DIFF
--- a/app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
+++ b/app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
@@ -440,13 +440,16 @@ class ChatRoomFragment : Fragment(), ChatRoomView, EmojiKeyboardListener, EmojiR
             subscribeTextMessage()
             emojiKeyboardPopup = EmojiKeyboardPopup(activity!!, activity!!.findViewById(R.id.fragment_container))
             emojiKeyboardPopup.listener = this
+
             text_message.listener = object : ComposerEditText.ComposerEditTextListener {
                 override fun onKeyboardOpened() {
-                    if (recycler_view.isAtBottom()) {
+                    var index = 0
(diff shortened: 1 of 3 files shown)


shriakhilc:
Can someone review this?

shriakhilc:
Behaviour present in beta6. Closing PR.

PR #898, opened 2018-03-13
[NEW] Unpin message from PinnedMessageActivity

### LINKED ISSUE #852: [New] [UX] There should be an option to unpin the pinned messages from "Pinned Message" activity
**Rocket.Chat.Android version:**
versionCode 1009
versionName "2.0.0-dev7"

# Description
Option to unpin the messages pinned by you is not there in pinned message activity.

**Willing to work on it?** YES

divyanshub024:
@RocketChat/android

Closes #852

### Description
This PR allows you to unpin the pinned Message from PinnedMessageActivity
OnLongClick on the pinned message the alertDialog will ask you to unpin the message.

device-2018-03-13-140448

### DIFF
--- a/app/src/main/java/chat/rocket/android/chatroom/adapter/BaseViewHolder.kt
+++ b/app/src/main/java/chat/rocket/android/chatroom/adapter/BaseViewHolder.kt
@@ -73,7 +73,7 @@ abstract class BaseViewHolder<T : BaseViewModel<*>>(
     interface ActionsListener {
         fun isActionsEnabled(): Boolean
         fun onActionSelected(item: MenuItem, message: Message)
-        fun onPinMessageSelected(message: Message)
+        fun onPinMessageSelected(item: MenuItem, message: Message)
     }
 
     val longClickListener = { view: View ->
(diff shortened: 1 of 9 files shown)


rafaelks:
@divyanshub024 Thank you for your PR! I believe we can show all the actions we're showing today on the messages list, instead of just this alert. What do you think?

divyanshub024:
@rafaelks I didn't get it. By this, you mean that every message should have an icon to unpin?

divyanshub024:
I also want to know why the circleci test failed?

mmarif4u:
I guess what @rafaelks means is the context menu appear by long holding on the message which have different actions like Delete, pin message etc etc.

rafaelks:
@divyanshub024 Exactly what @mmarif4u said.

divyanshub024:
@rafaelks @mmarif4u In PinnedMessageActivity you don't have actions like delete, pin message etc. PinnedMessageActivty only shows pinned messages.

divyanshub024:
@rafaelks How do you suggest this action to be performed?

mmarif4u:
2 ways to do it, one is to switch the action to either "Pin Message" if it not yet or "Unpin Message" if it is pinned already.

2nd is to leave it as it is what you have implemented so far.

I will leave it to the RC Core team to decide on that.

mmarif4u:
@divyanshub024 you are right, the pinned messages list is different from context menu.

divyanshub024:
@mmarif4u should I leave it for now or should I change it?

rafaelks:
@divyanshub024 You could still implement it... I could reply, copy, quote, unpin and delete a pinned message from the pinned list. In case of reply/quote, you could close the window and call the reply/quote action into the chat activity. What do you think?

divyanshub024:
@rafaelks I think its a great idea. I'll start working on it.

divyanshub024:
clone: #1029

PR #906, opened 2018-03-13
[FIX] Rounded avatars

philipbrito:
This PR removes the SVG support and requests the avatar images as jpeg or png files.


app/src/main/java/chat/rocket/android/authentication/signup/ui/SignupFragment.kt
```diff
@@ -19,12 +19,15 @@ import javax.inject.Inject
...
-
     private val layoutListener = ViewTreeObserver.OnGlobalLayoutListener {
-        if (KeyboardHelper.isSoftKeyboardShown(constraint_layout.rootView)) {
-            text_new_user_agreement.setVisible(false)
+        if (KeyboardHelper.isSoftKeyboardShown(relative_layout.rootView)) {
+            bottom_container.setVisible(false)
         } else {
-            text_new_user_agreement.setVisible(true)
+            bottom_container.apply {
+                postDelayed({
```
luciofm: postDelayed 3ms is a very big code smell. It would be the same as post or postOnAnimation

PR #929, opened 2018-03-15
[CI] Upgrade Circle CI to version 2.0

### LINKED ISSUE #821: [CI] Migrate CircleCI to v2
We have to migrate our CircleCI to version 2. They will shutdown v1 soon.

YusukeIwaki:
@RocketChat/android

Closes #821

Circle CI 2.0 works much faster than Circle CI 1.0.
Also we can separate the code-analysis job (checkstyle, pmd, ...) from the main build job with Circle CI 2.0, and run them in parallel like this 😆 

The new config.yml introduced in this PR doesn't include deployment processes (uploading APKs to Google Play alpha or beta) because `Triple-T/gradle-play-publisher` is not configured at this time.
(Also, I think it would be better to use [fastlane](https://github.com/fastlane/fastlane) for deploying APKs)


rafaelks:
@YusukeIwaki This is looking good to me, and I also agree on using fastlane! Thank you!!! 👏

PR #949, opened 2018-03-21
[FIX] Properly mark a room as read

### LINKED ISSUE #900: [BUG] New messages aren't being set as "read" when they arrive
Your Rocket.Chat.Android version: 2.0.0
Mobile device model and OS version: Sony Xperia

When I receive new messages and I'm inside of the chat, the chat isn't marking the messages as read. It should mark them as read.

This video shows the bug happening:

shriakhilc:
@RocketChat/android

Closes #900

Currently, the `markRoomAsRead()` function is called just once before displaying the messages. This messes up the unread marker display, so it has  been shifted to below the `loadMessages()` call.

New messages after the user joining the room are handled by `subscribeMessages()` function, and this subscription is removed when the user exits via the `unsubscribeMessages()` function. Since the method marks the entire room as read and not on a per message basis, a call to `markRoomAsRead()` has been added to `unsubscribeMessages()`. This seems to have fixed the issue.

PS: I am assuming it is desirable to mark the entire room as read when the user views and exits the room. There seems to be a `TODO` comment about marking only when the first unread message is read, and not otherwise. If that is the case, a slightly different approach will be needed.

### DIFF
--- a/app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomPresenter.kt
+++ b/app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomPresenter.kt
@@ -71,12 +71,12 @@ class ChatRoomPresenter @Inject constructor(private val view: ChatRoomView,
                         client.messages(chatRoomId, roomTypeOf(chatRoomType), offset, 30).result
                 messagesRepository.saveAll(messages)
 
+                val messagesViewModels = mapper.map(messages)
+                view.showMessages(messagesViewModels)
+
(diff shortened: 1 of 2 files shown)


shriakhilc:
@leonardoaramaki  @rafaelks please review

philipbrito:
Looking at some competitors for instance, they are marking a room as read when the user sees the first unread message. We can do that plus adding an "option" to mark the room whitout the need of seeing all unread messages (image a room with hundred of unread message).

philipbrito (approved):
Thank you!

PR #957, opened 2018-03-23
[FIX] Fixed Fragment Replacement in ChatRoomActivity

### LINKED ISSUE #956: [BUG] Fragment replaced during Orientation Change
Your Rocket.Chat.Android version: 2.0.0 (latest branch

Mobile device model and OS version: Nexus 6 - 7.0.1

**Information**
On orientation change, activities are destroyed and recreated meaning the `onCreate` method is called every orientation change. This is problematic for `ChatRoomActivity` which [replaces a fragment](https://github.com/RocketChat/Rocket.Chat.Android/blob/develop-2.x/app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomActivity.kt#L89) every time the orientation changes. When the back button, the activity will go the back stack and try to add the original `ChatRoomFragment` instance (from when the activity was created the first time) over the new one. This crashes the program since the fragment is already on the screen.

**Images**

samrmur:
@RocketChat/android

Closes #956 

**Information**
Fixed the issue by only adding the fragment if the fragment manager cannot find it within its back-stack

### DIFF
--- a/app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomActivity.kt
+++ b/app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomActivity.kt
@@ -86,9 +86,11 @@ class ChatRoomActivity : AppCompatActivity(), HasSupportFragmentInjector {
 
         isChatRoomSubscribed = intent.getBooleanExtra(INTENT_CHAT_IS_SUBSCRIBED, true)
 
-        addFragment("ChatRoomFragment", R.id.fragment_container) {
-            newInstance(chatRoomId, chatRoomName, chatRoomType, isChatRoomReadOnly, chatRoomLastSeen,
-                    isChatRoomSubscribed)
+        if (supportFragmentManager.findFragmentByTag("ChatRoomFragment") == null) {
+            addFragment("ChatRoomFragment", R.id.fragment_container) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomActivity.kt
```diff
@@ -86,9 +86,11 @@ class ChatRoomActivity : AppCompatActivity(), HasSupportFragmentInjector {
 
         isChatRoomSubscribed = intent.getBooleanExtra(INTENT_CHAT_IS_SUBSCRIBED, true)
 
-        addFragment("ChatRoomFragment", R.id.fragment_container) {
-            newInstance(chatRoomId, chatRoomName, chatRoomType, isChatRoomReadOnly, chatRoomLastSeen,
-                    isChatRoomSubscribed)
+        if (supportFragmentManager.findFragmentByTag("ChatRoomFragment") == null) {
```
philipbrito, line 89: Next time we'll need to add a `const` with the tag for our fragments.

philipbrito (approved):
Thank you!

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomActivity.kt
samrmur: Thanks for letting me know, I will remember for next time.

PR #967, opened 2018-03-28
[FIX] Web preview (Custom tabs).

ear-dev:
@RocketChat/android

This fixes two bugs mentioned in PR 917.
 
1) Clicking web preview links resulted in opening a webView and not a custom tab.  It now opens a custom tab. 
2) A capital 'H' in Http://www.**.com urls would cause the app to crash.  This PR converts to lower case before passing to custom tabs.
3) The PR also contains a small refactor to implement a dependency correctly.


app/build.gradle
```diff
@@ -111,7 +111,7 @@ dependencies {
         exclude group: 'com.android.support', module: 'support-annotations'
     })
 
-    implementation 'com.android.support:customtabs:27.0.2'
+    implementation libraries.customTabs
```
philipbrito: Thanks for this change.

dependencies.gradle
```diff
@@ -30,6 +30,7 @@ ext {
         sheetMenu              : '1.3.3',
         aVLoadingIndicatorView : '2.1.3',
         flexbox                : '0.3.2',
+        customTabs             : '27.0.2',
```
filipedelimabrito: Thanks again.

dependencies.gradle
```diff
@@ -94,6 +95,8 @@ ext {
 
         aVLoadingIndicatorView : "com.wang.avi:library:${versions.aVLoadingIndicatorView}",
 
+        customTabs             : "com.android.support:customtabs:${versions.customTabs}",
```
philipbrito: Thanks again.

app/src/main/java/chat/rocket/android/chatroom/adapter/UrlPreviewViewHolder.kt
```diff
@@ -32,8 +32,11 @@ class UrlPreviewViewHolder(itemView: View,
             text_title.content = data.title
             text_description.content = data.description ?: ""
 
+            val tabsbuilder = CustomTabsIntent.Builder()
+            val customTabsIntent = tabsbuilder.build()
+
             url_preview_layout.setOnClickListener { view ->
-                view.context.startActivity(Intent(Intent.ACTION_VIEW, Uri.parse(data.rawData.url)))
+                customTabsIntent.launchUrl(context, Uri.parse(data.rawData.url.toLowerCase()))
```
philipbrito: You can use `data.rawData.url.toLowerCase().toUri`.

Fresh from Android KTX -> https://android.github.io/android-ktx/core-ktx/androidx.net/kotlin.-string/to-uri.html.

philipbrito (approved):
Thanks!

dependencies.gradle
philipbrito: You don't need to specify it here. CustomTabs uses Support Library, so remove it and use 
https://github.com/RocketChat/Rocket.Chat.Android/blob/ba95c1abb4b0b08d22f8f7cf8a2f6019df428eac/dependencies.gradle#L12

philipbrito (changes requested):
Only a small change and thanks!

Shailesh351:
@ear-dev I don't think converting url to lower case is good fix, it can break url. Some url have uniqe ids which have uppercase letters also(like google drive sharing link)

leonardoaramaki:
@Shailesh351 is right. You should do `HttpUrl.parse().host().toLowerCase()` on the host only and then append the rest of the string. It's a very weird bug on custom tabs. What you could also do is  `HttpUrl.parse().schema().toLowerCase()`. Haven't tested but I think what's breaking is just the schema. Then you concatenate the rest of the url. `HttpUrl` can help you with that.

philipbrito (approved):

philipbrito:
@ear-dev Just do the changes requested/appointed by @Shailesh351 and @leonardoaramaki and done.

ear-dev:
@Shailesh351 @filipedelimabrito @leonardoaramaki It took me a while to figure out how to use HttpUrl, not to mention that I am also new to Java, but I have a fix for only lower casing the scheme.  I put the method in a helper file and I just guessed as to a relevant location for that file.  I tested all scenarios that we know about, so this worked.  Thanks!

leonardoaramaki:
@ear-dev Can you please solve the conflicts?

ear-dev:
done.. thanks @leonardoaramaki

rafaelks:
@ear-dev Thanks for the PR! Just waiting for the conflicts to be solved in order to merge it.

rafaelks:
This bug is going to be handled by @luciofm in another PR. Thanks for the PR @ear-dev! 👍

rafaelks:
This one was fixed by #1548. Thanks @ear-dev!

PR #986, opened 2018-03-30
[FIX] Issues with OAuth.

philipbrito:
This PR fixes some issues with OAuth.


rafaelks:
@filipedelimabrito GitHub still have the same problem. Google works. GitLab crashes. LinkedIn button works, but after logging in it stays in a blank page (I think the problem is on their side now)

rafaelks (changes requested):
Still needs work.

philipbrito:
@rafaelks Needs to figure out what is going on. I made testes here and I don't get the above issues. Check a demo for GitHub for example: https://youtu.be/DPxT9BNPxJo

Could you test it in another device and/or install the app again?

rafaelks:
@filipedelimabrito Now GitHub is working for me. BUT if the user is not found/not registered, the app crashes in all situations. We need to fix that also in another PR.

rafaelks (approved):

Issue #985, opened 2018-03-30
[URGENT][BUG] Tapping on "Members" list navigates to the right with the icon of the room


rafaelks:
@leonardoaramaki Let's just do the same as we do on "Pinning message"

leonardoaramaki:
@rafaelks Shouldn't we just remove/hide that icon? I think that will suffice. The sliding to the right is pretty intuitive. :)

rafaelks:
Closing this one.

PR #991, opened 2018-03-31
[FIX] Remove Facebook, Twitter and Meteor auth methods

### LINKED ISSUE #982: [URGENT][IMPROVEMENT] Remove Facebook, Twitter and Meteor buttons
@leonardoaramaki We're not supporting this authentication methods right now, so let's just remove it from the screen so we don't get users confused.

leonardoaramaki:
@RocketChat/android

Also changed icon when swapping to members list from chatrrom (see #985).

Closes #982 and #985


rafaelks (approved):
👍

PR #1024, opened 2018-04-05
[IMPROVEMENT] Keep message on the text composer on fail to send message

### LINKED ISSUE #1007: [IMPROVEMENT] Block TextField until message is sent as a temporary solution for not using database

leonardoaramaki:
@RocketChat/android

I'd rephrase the #1007 issue though not to block the composer but instead just keep the message in there until a successful message sending.

Closes #1007


rafaelks (approved):
Looking good to me! 👍

PR #1026, opened 2018-04-05
[NEW] Add the possibility to change own status.

philipbrito:
**NOTE:** This is not listening to status changes from "external" sources (e.g.: Change the status from the web app to show it on the app), only vice versa.


rafaelks (approved):
Feature is looking great to me! 👏

PR #1034, opened 2018-04-06
[IMPROVEMENT] Get own current status (REST).

philipbrito:
This PR get the correct/current own status when opening the nav. drawer.


app/src/main/java/chat/rocket/android/app/DrawableHelper.kt
```diff
@@ -106,19 +106,21 @@ object DrawableHelper {
...
+            is UserStatus.Busy -> {
+                getDrawableFromId(R.drawable.ic_status_busy_24dp, context)
+            }
+            else -> getDrawableFromId(R.drawable.ic_status_invisible_24dp, context)
         }
-        return userStatusDrawable
     }
 
-    // TODO Why we need to UserStatus?
+    // TODO Why we need two UserStatus?
```
Shailesh351, line 123: https://github.com/Konecty/meteor-user-presence

Connection status is used to set temporary status which is used while app is open, open in background.
We can use it as 
App is running -> Connection Status(online)
App is running in background or closed ->  Connection Status(away)

philipbrito: Right! But we have two UserStatus class in the SDK that "does" the same thing. It is the question.

PR #1073, opened 2018-04-11
[IMPROVEMENT] Offline message sending support

### LINKED ISSUE #1070: [IMPROVEMENT] Improve the "Send message" behaviour
Your Rocket.Chat.Android version: 2.0.0-beta6

# Description

- [x] Message needs to be displayed as temporary on the list of message (light gray color);
- [x] Text field needs to be cleared on tapping "Send";
- [x] User needs to be able to send multiple messages in sequence without needing to wait for any message to complete the request;
- [ ] In case of an error, show the "exclamation" icon in red with the ability to re-send the message;

leonardoaramaki:
@RocketChat/android

This PR needs https://github.com/RocketChat/Rocket.Chat.Kotlin.SDK/pull/128 to be merged first.

Closes #1070


app/src/main/java/chat/rocket/android/chatroom/adapter/MessageViewHolder.kt
```diff
@@ -8,9 +9,9 @@ import kotlinx.android.synthetic.main.avatar.view.*
 import kotlinx.android.synthetic.main.item_message.view.*
 
 class MessageViewHolder(
-    itemView: View,
-    listener: ActionsListener,
-    reactionListener: EmojiReactionListener? = null
+        itemView: View,
```
luciofm: As per [Kotlin Coding Conventions](https://kotlinlang.org/docs/reference/coding-conventions.html#class-header-formatting), the old indentation was correct.

leonardoaramaki: They said it's 4 spaces. I'll change my tab spacing on AS to 4.

app/src/main/java/chat/rocket/android/chatroom/adapter/MessageViewHolder.kt
```diff
@@ -29,6 +30,9 @@ class MessageViewHolder(
             text_sender.text = data.senderName
             text_content.text = data.content
             image_avatar.setImageURI(data.avatar)
+            text_content.setTextColor(
```
luciofm, line 33: Maybe we could use a `ForegroundColorSpan` on the ViewModel to keep all the formating there, but really don't know which one is better/cheaper: `setTextColor` or using a Span

app/src/main/java/chat/rocket/android/dagger/module/AppModule.kt
```diff
@@ -278,4 +287,22 @@ class AppModule {
             getSettingsInteractor: GetSettingsInteractor): PushManager {
         return PushManager(groupedPushes, manager, moshi, getAccountInteractor, getSettingsInteractor, context)
     }
+
+    @Provides
+    fun provideJobScheduler(context: Application): JobScheduler {
+        return context.getSystemService(Context.JOB_SCHEDULER_SERVICE) as JobScheduler
```
luciofm, line 293: `return context.systemService()` and voilà

leonardoaramaki: We can't use that because this call requires API 23.

app/src/main/java/chat/rocket/android/chatroom/adapter/MessageViewHolder.kt
leonardoaramaki: That's a good point. I don't know which one would be most performant but keeping all the formatting logic separated is certainly better.

rafaelks:
Looking VERY good @leonardoaramaki!

PR #1074, opened 2018-04-11
[FIX] always run view callbacks on ui thread

luciofm:
Also check for `activity` and `view` not null to avoid NPEs


rafaelks:
Looking good! 👍

PR #1093, opened 2018-04-16
[NEW] Auth deep link

luciofm:
Add support for auth deeplink

[https://rocket.chat/docs/developer-guides/deeplink/]
Supports both rocketchat://auth and https://go.rocket.chat/auth urls


rafaelks:
Looking great! 👍

PR #1126, opened 2018-04-19
[NEW] Facebook oauth login

luciofm:


app/src/main/java/chat/rocket/android/helper/OauthHelper.kt
```diff
@@ -67,4 +67,21 @@ object OauthHelper {
                 "&response_type=code" +
                 "&scope=read_user"
     }
+
+    /**
+     * Returns the Facebook Oauth URL.
+     *
+     * @param clientId The Gitlab client ID.
```
philipbrito: It should be `The Facebook client ID`.

philipbrito (changes requested):

rafaelks:
Looking good!!! 👏

PR #1141, opened 2018-04-20
Remove the request for READ_EXTERNAL_STORAGE permission

MariusVolkhart:
We use the Storage Access Framework (SAF) to enable users to send files. When using the SAF, READ_EXTERNAL_STORAGE permission is not needed to read from internal storage, SD Cards, or anything else. The Android System is the one accessing the storage, so applications do not need direct access. This is in fact one of the major selling points of the SAF.

Note that no Google samples or apps by Google Developer Advocates that highlight the SAF request this permission.

Removing this permission is a good thing. Firstly, there are many SAF providers such as Google Drive, Photos, Microsoft OneDrive, etc, that do not use external storage. Secondly, users are not prompted for a needless permission. Thirdly, it reduces the complexity of the ChatRoomFragment.

@RocketChat/android

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -5,7 +5,6 @@
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission android:name="android.permission.VIBRATE" />
-    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
     <uses-permission android:name="com.google.android.c2dm.permission.RECEIVE" />
 
     <permission
--- a/app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
+++ b/app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
@@ -1,18 +1,14 @@
 package chat.rocket.android.chatroom.ui
 
-import android.Manifest
 import android.app.Activity
 import android.content.ClipData
 import android.content.ClipboardManager
 import android.content.Intent
-import android.content.pm.PackageManager
 import android.net.Uri
 import android.os.Bundle
 import android.os.Handler
 import android.support.annotation.DrawableRes
-import android.support.v4.app.ActivityCompat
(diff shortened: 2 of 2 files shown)


philipbrito:
Very good explanation! Thank you Sir!

leonardoaramaki:
@MariusVolkhart @filipedelimabrito I don't quite remember now but I put this check for permission there and I think we will need this later to fetch metadata from files prior to uploading. This will be needed to render the preview later on.

MariusVolkhart:
Hey @leonardoaramaki 

I'm not sure I follow, but I definitely want to address any issues with the change.

I tested this on an emulator running API 27. After removing the permission I sent 3 files (2 pdfs and a jpeg) to a friend of mine; one was from the SD Card, one from the Downloads provider, and one from internal storage. The files all get sent correctly (as in, I can see them on another device) and the image renders a preview (PDFs do not). Just to be safe, I just tried with several more providers (Google Drive, Google Photos, a demo app from the Android samples, and a provider I've implemented).

The following is a screenshot of the conversation after I had removed the permission and sent the first 3 files using SAF.
screenshot_1524255697

The [developer docs](https://developer.android.com/guide/topics/providers/document-provider.html#permissions) do indicate that the URI permissions will be revoked after a reboot. However, after restarting my device, the previews still render (my guess is that they are loaded from the data in Rocket Chat, not the original URI, but I didn't confirm).

Anyhow, again, I definitely want to address any issues with the change, but at this time, I do not believe the permission is necessary.

leonardoaramaki:
@MariusVolkhart Actually it was regarding video previews, which was being handled on a separate branch I was working on. But it was just an experiment at the time and the branch doesn't exist anymore (on the roadmap for later).
I don't remember exactly what was the exact point I got the error but it was a silent one with no crashes whatsoever. 
I remember there was also some issue with code to get the file size which was returning innacurate, don't know if it's related. 
Indeed for uploading files using the current method we don't need this permission and your change is accurate.
I'll go through this again when I get to the upload preview and progress next time, then if I happen to get the error again I'll tag you so you could follow up if you want.
Thanks a lot for your time looking into all this. :relaxed:

PR #1172, opened 2018-04-24
[CHORE] Remove the Travis configuration

MariusVolkhart:
Travis isn't being used. I found this confusing as I was going to re-enable unit tests, but then realized CircleCI is used instead.

@RocketChat/android


philipbrito (approved):
Thank you @MariusVolkhart !

PR #1204, opened 2018-04-26
[FIX] Use username when building replies always

### LINKED ISSUE #1199: [BUG] Replying/Quoting messages using real name doesn't work
The URL needs to use the username to be constructed, instead of the full name.

### LINKED ISSUE #1200: [BUG] On autocomplete, use "username" instead of "real name" even if the setting is configured to use real name in all the places
The API will only understand the mention if the app sends username.

leonardoaramaki:
@RocketChat/android

Closes #1199 
Closes #1200

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -104,7 +104,6 @@ dependencies {
     implementation libraries.frescoImageViewer
 
     implementation libraries.markwon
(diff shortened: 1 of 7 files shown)


rafaelks:
Looking good @leonardoaramaki 👍

PR #1212, opened 2018-04-28
[IMPROVEMENT] Change ImagePipelineConfig initialization in Dagger graph

MariusVolkhart:
Rather than creating a new, unsized HashSet, use idiomatic Kotlin to create the set of RequestListeners for the ImagePipelineConfig. Under the hood a SingletonSet will be used which is optimized for the current use case, but if additional Listeners are added, the collections will automatically be sized appropriately. This leads to better memory utilization and reduced CPU usage.

@RocketChat/android


philipbrito (approved):
Good finding! Good explanation! Good PR!
Thank you @MariusVolkhart!

PR #1213, opened 2018-04-28
[IMPROVEMENT] Remove the FrescoAuthInterceptor

MariusVolkhart:
The FrescoAuthInterceptor has not been used since it was introduced, suggesting that it is not required for proper function. By not needing the FrescoAuthInterceptor, it is not necessary to copy or store the OkHttpClient.

Removing the FrescoAuthInterceptor produces a smaller APK due to there being less code. It reduces app cold start times by reducing the CPU time spent initializing an unused class and copying the OkHttpClient. It reduces RAM consumption by not keeping a deep copy of the OkHttpClient around for the lifetime of the application. Finally, it reduces compile time as less time is spent performing Dagger graph analysis.

@RocketChat/android


philipbrito (approved):
Good catch and great explanation!
Really thank you for your contribution!

philipbrito:
@MariusVolkhart Could you fix your PR? It isn't building.

MariusVolkhart:
@filipedelimabrito rebased onto `develop`. Build failure appears to stem from the fact that is was built from my fork instead of the main repo, so signing keys are missing

PR #1214, opened 2018-04-28
[IMPROVEMENT] Change OkHttpClient initialization in Dagger graph

MariusVolkhart:
Rather than using the Kotlin apply function, directly invoke the builder methods on the OkHttpClient.Builder instance. This produces noticeably better bytecode and the apply function does not add value in this case.

@RocketChat/android


philipbrito (approved):
Really thank you @MariusVolkhart!

PR #1220, opened 2018-04-29
[FIX] File uploading when no filters are applied

MariusVolkhart:
When no file filters are specified in the server whitelist, the Android system is incorrectly given a mime type filter list including a single empty ("") mime type. Naturally, nothing matches this, and no documents can be selected for upload.

To remedy this, the application is changed to return a null filter array if filters are not to be applied.

@RocketChat/android

### DIFF
--- a/app/src/main/java/chat/rocket/android/server/domain/SettingsRepository.kt
+++ b/app/src/main/java/chat/rocket/android/server/domain/SettingsRepository.kt
@@ -1,6 +1,5 @@
 package chat.rocket.android.server.domain
 
-import chat.rocket.android.util.extensions.mapToTypedArray
 import chat.rocket.core.model.Value
 
 typealias PublicSettings = Map<String, Value<Any>>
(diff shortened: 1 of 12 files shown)


app/src/main/java/chat/rocket/android/server/domain/SettingsRepository.kt
```diff
@@ -84,13 +83,16 @@ fun PublicSettings.allowedMessagePinning(): Boolean = this[ALLOW_MESSAGE_PINNING
...
-fun PublicSettings.uploadMimeTypeFilter(): Array<String> {
-    val values = this[UPLOAD_WHITELIST_MIMETYPES]?.value
-    values?.let { it as String }?.split(",")?.let {
-        return it.mapToTypedArray { it.trim() }
+fun PublicSettings.uploadMimeTypeFilter(): Array<String>? {
+    val values = this[UPLOAD_WHITELIST_MIMETYPES]?.value as String?
+    if (!values.isNullOrBlank()){
+        // It's far more efficient to split the string to an array using the Java type rather than the Kotlin type
+        @Suppress("PLATFORM_CLASS_MAPPED_TO_KOTLIN")
+        val filter = (values as java.lang.String).split(",")
```
luciofm, line 91: On this case I'd prefer readability over performance. How much difference this will make here? Doing a simple measure, the gain seems negligible.
`return values.split(",").mapToTypedArray { it.trim() }` is more readable and idiomatic.

app/src/main/res/values/defaults.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <string name="default_protocol" translatable="false">https://</string>
```
luciofm, line 3: Don't change the default protocol.

luciofm (changes requested):
Thanks for the PR, just some comments

luciofm:
Applied manually on beta...

PR #1235, opened 2018-05-02
[FIX] Fix sorting on chatrooms

luciofm:
Also check for STORE_LAST_MESSAGE or refresh settings

### DIFF
--- a/app/src/main/java/chat/rocket/android/chatrooms/presentation/ChatRoomsPresenter.kt
+++ b/app/src/main/java/chat/rocket/android/chatrooms/presentation/ChatRoomsPresenter.kt
@@ -64,6 +64,10 @@ class ChatRoomsPresenter @Inject constructor(
             view.showLoading()
             subscribeStatusChange()
             try {
+                // If we still don't have 'Store_Last_Message' setting, refresh the settings
+                if (!settings.hasShowLastMessage()) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/chat/rocket/android/chatrooms/presentation/ChatRoomsPresenter.kt
```diff
@@ -238,6 +244,14 @@ class ChatRoomsPresenter @Inject constructor(
         }
     }
 
+    private fun chatRoomTimestamp(chatRoom: ChatRoom): Long? {
+        return if (settings.hasShowLastMessage() && settings.showLastMessage()) {
+            chatRoom.lastMessage?.timestamp ?: chatRoom.updatedAt
```
luciofm, line 249: @rafaelks is this OK? if a room doesn't have last message recorded yet, we will use the `updatedAt` for this room

rafaelks (approved):

app/src/main/java/chat/rocket/android/chatrooms/presentation/ChatRoomsPresenter.kt
rafaelks: @luciofm Yes, perfect! 👌

Issue #116, opened 2016-12-26
[NEW] Support for messaging typing status


srinivas3120:
- [x] Sending typing status.
- [x] Listening to typing changes.

alexfdz:
@rafaelks @filipedelimabrito do you have any timeline for this feature?

rafaelks:
@alexfdz This is not our highest priority right now for internal development, but if someone is willing to send us a PR it would be super welcome! If there's no PR, I would say that version 2.2.0 (end of May) will have it.

alexfdz:
thanks for your reply @rafaelks! Is unlikely I will have time before 2.2.0, but probably I could help with some testing if you need a hand.

PR #1238, opened 2018-05-02
[NEW] Send typing status feature.

philipbrito:
Depends on https://github.com/RocketChat/Rocket.Chat.Kotlin.SDK/pull/149

Closes partly #116


app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
```diff
@@ -674,7 +674,10 @@ class ChatRoomFragment : Fragment(), ChatRoomView, EmojiKeyboardListener, EmojiR
 
     private fun subscribeTextMessage() {
         val disposable = text_message.asObservable(0)
-            .subscribe({ t -> setupComposeMessageButtons(t) })
+            .subscribe { t ->
+                setupComposeMessageButtons(t)
+                sendTypingStatus(t)
```
rafaelks: Is this being called on every change of the text field? If yes, we need to add a "buffer" of ~3 seconds on each call to avoid many calls repeated.

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
```diff
@@ -697,6 +700,14 @@ class ChatRoomFragment : Fragment(), ChatRoomView, EmojiKeyboardListener, EmojiR
         }
     }
 
+    private fun sendTypingStatus(charSequence: CharSequence) {
+        if (charSequence.isNotEmpty()) {
+            presenter.sendTyping()
```
rafaelks: What happens if I remove a char?

filipedelimabrito: You need to take a look in how this is working. If you remove a `char` from the `charSequence` and the `charSequence`  is not empty (see `charSequence.isNotEmpty()`) it means that there are char in the composer otherwise there is no char in the composer, so called `presenter.sendNotTyping`.

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
philipbrito: You need to take a look in how this is working. It is how it was prior to this PR. Note that we need to set up the composer message buttons too.

It is better to see it working. 
Could you say if you see any app delay, more process when using this feature?

rafaelks: @filipedelimabrito Yes, I see it working and I saw how this is working. You're sending one event to the WebSocket on every change, so if you're a message with 5000 chars, there will be 5000 calls on the WebSocket. This is causing flickering on the web application and is not necessary.

You need to add a buffer, and only call the WebSocket every 3 seconds, instead of every char.

philipbrito: Ohh, I see. You're right. I was just trying to replicate the typing behaviour of some apps like Telegram. The bad with that is: If the user type as fast he/she can, we'll never send the typing status, and it could be more bad if we think about the possibility of the user using the **handwriting**.

I made tests using the Telegram mobile app and desktop app (just for reference) and it is smart enough to change the typing status, independently if I'm typing super fast in the mobile/desktop application.

But yeah, for our websocket I've made changes.
Thanks for clarifying.

PR #1250, opened 2018-05-07
[IMPROVEMENT] Change password layout

### LINKED ISSUE #1097: [IMPROVEMENT] Change password layout
Presently in change password user will see the done/save icon in action bar to save new password only when the user enters the correct password in "confirm new password", which should not be the case as it would confuse them how to change the password in case of the wrong password.

It is better to show the button always and then show error/alert to enter the correct password.  

ps: we should also ask for current password.

divyanshub024:
@RocketChat/android

Closes #1097


divyanshub024:
Closing since that layout will be change very soon

filipedelimabrito:
*Closed since that layout will be change very soon*.

PR #1258, opened 2018-05-07
[NEW] Detect redirect on server info

luciofm:
Also ignore malformed sendMessage response when sending on a readonly channel


rafaelks:
Looking good to me! 👍

rafaelks (approved):

Issue #26, opened 2016-07-20
[REQUEST] Upload the app to F-DROID


tpeg5stan:
Hello,

As the licence (MIT) is compatible with the FOSS catalog "F-DROID", is it possible to get the app there ?
I am a Rocket.Chat user on my desktop browser, but on my mobile browser it is not easy to use. Moreover, I have uninstalled Google Play Store, and thus won't get the app on it.

Thanks in advance

geekgonecrazy:
@tpeg5stan I don't believe this app is feature complete enough to be listed in any store.  But... please see this issue https://github.com/RocketChat/Rocket.Chat.Cordova/issues/104 for our main app

tpeg5stan:
Oh, I did not understand that, you have 3 different apps for Android ? Which one should I use (at the moment, I prefer using my browser) ? Is one of the 3 projects abandoned ?
Don't worry about the fact that the app is not finished, there are beta and even alpha apps on F-DROID, I am a volunteer to test it if needed :)

geekgonecrazy:
The Rocket.Chat.Cordova is most actively maintained, and currently the only one we as the core team are maintaining.  As soon as we have the bandwidth we'll be venturing into the Native apps, and at that time we will probably combine the two projects if need be.

herbsmn:
fyi @tpeg5stan: https://github.com/RocketChat/Rocket.Chat.Cordova/issues/104#issuecomment-248840584
I think it still uses Google for push notifications though, so it can't be added to the regular f-droid FOSS repo yet.

tpeg5stan:
Oh, yes, I see that this thread has parallel issues.
I perfectly understand the problem, it is difficult ot take away all the google specific requirements, but once it is done, app is 100%open-source.
Is it possible to have an application without notification yet ? Like an alpha-version or so...

geekgonecrazy:
Lets continue conversation over here: RocketChat/Rocket.Chat.Cordova#104.  Closing this one.

schaengel:
Since Cordova app looks like its unmaintained. The conversation of creating FDroid app belongs here.

geekgonecrazy:
Situation has definitely changed since I closed this.  The other is still maintained, but this one is much more active

rafaelks:
👍

Bubu:
We have a request for packaging issue now in the f-droid tracker: https://gitlab.com/fdroid/rfp/issues/436

[...]

PR #1290, opened 2018-05-20
[IMPROVEMENT] Update gradle-wrapper.properties by adding SHA-256 comparison

Poussinou:
@RocketChat/android

Hi,

This small PR adds the `distributionSha256Sum` property to `gradle-wrapper.properties`. It allows for verification of the downloaded Gradle distribution via SHA-256 hash sum comparison.

You can find some documentation about this security improvement [here](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:verification), and the sums for each version of gradle [here](https://services.gradle.org/distributions/).

Note that you need to update this property every time you change your version of gradle.

#26


philipbrito (approved):
Thanks for this PR (and for your explanation) @Poussinou!

Poussinou:
No problem 👍

PR #1331, opened 2018-06-01
[FIX] Use named arguments to avoid overriding the password on profile updating

### LINKED ISSUE #1308: [BUG] Lost account after change name
## Description

Lost account after change display name of my user in App

## Devices and Versions

Your Rocket.Chat.Android version: 2.2.0
Your Rocket.Chat Server version: 0.64.2

Mobile device model and OS version: Sansung J7 Android 5.1.1

## Steps to reproduce

I installed a rocketchat using docket. After that I accessed it using the Android App, I logged into the user profile and changed my display name, not my username, he said he had changed successfully and then he said he should be logged in.
After this I could no longer access my rocketchat, the function of recovering the password did not work either, whenever I tried to change it informed that there was an error, and even trying to change the password directly in mongodb also had no result.
To confirm the problem I removed the docker installation, deleted the database on the mongo and installed it again.
I tried changing the display name inside the web panel and it worked without problems, then I did the process again using the Android App and totally lost the access again.
I started testing rocketchat 2 days ago so I still do not know the project enough to give more details of what may be causing this problem.

leonardoaramaki:
@RocketChat/android

Closes #1308

### DIFF
--- a/app/src/main/java/chat/rocket/android/profile/presentation/ProfilePresenter.kt
+++ b/app/src/main/java/chat/rocket/android/profile/presentation/ProfilePresenter.kt
@@ -58,7 +58,8 @@ class ProfilePresenter @Inject constructor(private val view: ProfileView,
                 if(avatarUrl!="") {
                     retryIO { client.setAvatar(avatarUrl) }
                 }
-                val user = retryIO { client.updateProfile(myselfId, email, name, username) }
+                val user = retryIO { client.updateProfile(
+                    userId = myselfId, email = email, name = name, username = username) }
                 view.showProfileUpdateSuccessfullyMessage()
                 loadUserProfile()
             } catch (exception: RocketChatException) {


app/src/main/java/chat/rocket/android/profile/presentation/ProfilePresenter.kt
```diff
@@ -58,7 +58,8 @@ class ProfilePresenter @Inject constructor(private val view: ProfileView,
                 if(avatarUrl!="") {
                     retryIO { client.setAvatar(avatarUrl) }
                 }
-                val user = retryIO { client.updateProfile(myselfId, email, name, username) }
+                val user = retryIO { client.updateProfile(
+                    userId = myselfId, email = email, name = name, username = username) }
```
luciofm, line 62: 👍

PR #1336, opened 2018-06-02
[IMPROVEMENT] Show toast on copy message

### LINKED ISSUE #1335: [IMPROVEMENT] Show toast on copy message
## Description
It's better to show toast with message "Message copied"  on click of copy action.

divyanshub024:
@RocketChat/android

Closes #1335


rafaelks (approved):

app/src/main/java/chat/rocket/android/app/DateTimeHelper.kt
```diff
@@ -17,8 +17,8 @@ object DateTimeHelper {
      * @param long The [Long]
      * @return The [LocalDateTime] from a [Long].
      */
-    fun getLocalDateTime(long: Long): LocalDateTime {
-        return LocalDateTime.ofInstant(Instant.ofEpochMilli(long), ZoneId.systemDefault())
+    fun getLocalDateTime(long: Long?): LocalDateTime {
```
philipbrito, line 20: Why did you make this change? You are not using this function on this PR, so is there a specific case that needs the change here?

c/c @rafaelks

divyanshub024: After the latest merge it was showing a type mismatch error, required long? instead of long. That's why I changed it to long?.

philipbrito: It never should accept null since this function requires a non-null to work!

divyanshub024: I'm getting this error in FileViewModel line 45. 
```
/home/divyanshu/AndroidStudioProjects/Rocket.Chat.Android/app/src/main/java/chat/rocket/android/files/viewmodel/FileViewModel.kt: (45, 45): Type mismatch: inferred type is Long? but Long was expected
```

philipbrito: What I'm saying is that `getLocalDateTime()` is an end function that expects a Long to work, so why this function, as it is, should expect a null value? Kotlin is **safe-nullability** and we never should do this thing. 

Because you're getting an error in the `FileViewModel` class it would be better to call this function only when the long from that is not null, not the other way around, exposing the `getLocalDateTime()` to everyone that it receives null values, when it actually works only with non-null values.

I'll fix that.

Issue #1342, opened 2018-06-05
[IMPROVEMENT] Clear OAuth cache after a successful login 


philipbrito:
Right now, if an user do a successful login using any OAuth providers he/she will not be able to login with a different user in that provider (only clearing the app cache in the system settings).

PR #1344, opened 2018-06-05
[IMPROVEMENT] Removes web view cookies for OAuth and SSO.

philipbrito:
Closes https://github.com/RocketChat/Rocket.Chat.Android/issues/1342

### DIFF
--- a/app/src/main/java/chat/rocket/android/webview/oauth/ui/OauthWebViewActivity.kt
+++ b/app/src/main/java/chat/rocket/android/webview/oauth/ui/OauthWebViewActivity.kt
@@ -6,6 +6,7 @@ import android.content.Context
 import android.content.Intent
 import android.os.Bundle
 import android.support.v7.app.AppCompatActivity
+import android.webkit.CookieManager
 import android.webkit.WebView
 import android.webkit.WebViewClient
 import androidx.core.net.toUri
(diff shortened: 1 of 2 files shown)


rafaelks (approved):

app/src/main/java/chat/rocket/android/webview/oauth/ui/OauthWebViewActivity.kt
```diff
@@ -56,6 +57,11 @@ class OauthWebViewActivity : AppCompatActivity() {
         }
     }
 
+    override fun onDestroy() {
+        super.onDestroy()
+        CookieManager.getInstance().removeAllCookies(null)
```
luciofm: Wouldn't be better to clean it on `onCreate` before setting up the `WebView`, because if for some reason we crash after the user authenticated, while parsing the response, we will never cleanup the cookies.

app/src/main/java/chat/rocket/android/webview/sso/ui/SsoWebViewActivity.kt
```diff
@@ -52,6 +53,11 @@ class SsoWebViewActivity : AppCompatActivity() {
         }
     }
 
+    override fun onDestroy() {
+        super.onDestroy()
+        CookieManager.getInstance().removeAllCookies(null)
```
luciofm: Same as above

app/src/main/java/chat/rocket/android/webview/oauth/ui/OauthWebViewActivity.kt
philipbrito: @luciofm Changed.

luciofm (approved):
👍

PR #1351, opened 2018-06-08
[RELEASE] 2.3.2 - HOTFIX

luciofm:


app/src/main/java/chat/rocket/android/infrastructure/CrashlyticsWrapper.kt
```diff
@@ -0,0 +1,69 @@
...
+            if (crashlyticsHandler != null) {
+                crashlyticsHandler.uncaughtException(t, e)
+            } else {
+                throw RuntimeException("Missing default exception handler")
+            }
+        }
+    }
+}
+
+private const val KEY_CURRENT_SERVER = "CURRENT_SERVER"
```
philipbrito, line 61: Following the code style, this shouldn't be on the top (after `CrashlyticsWrapper`) ?

luciofm: Personally I prefer what matters first, in this case the code... This are just self descripting private constants.

philipbrito: Yep! I know your approach and I like it. But if we are following the [convention](https://kotlinlang.org/docs/reference/coding-conventions.html#class-layout) we always should respect it, otherwise we'll always have personal preferences here.

Because we are following that convection since the beginning my advice is to change it to avoid our codebase with multiple (personal) approach.

luciofm: I don't think it doesn't respect the convention.

# Source file organization
Placing multiple declarations (classes, top-level functions or properties) in the same Kotlin source file is encouraged as long as these declarations are closely related to each other semantically and the file size remains reasonable (not exceeding a few hundred lines).

In particular, when defining extension functions for a class which are relevant for all clients of this class, put them in the same file where the class itself is defined. When defining extension functions that make sense only for a specific client, put them next to the code of that client. Do not create files just to hold "all extensions of Foo".

# Class layout
Generally, the contents of a class is sorted in the following order:

* Property declarations and initializer blocks
* Secondary constructors
* Method declarations
* Companion object

First because the constants are top level, and not declared inside the class itslef, and as per the convention, they should be below the class.

Also I don't classify constants as properties declarations neither initializer blocks, they are more closely related to Companion objects, where you declare some helper functions.

So anything that is declared outside the class should come after it.

luciofm: And, there is no value to blindly follow a convention, we can always adapt it to our needs.... A convetion is this, just a convention, it is not a syntax rule :)

philipbrito: In my point of view, constants are always **properties**. But my point of view isn't the "king view" so I became more convicted of that, after reading this: 

*"Constants are val **properties** with no custom get function, whose contents are deeply immutable, and whose functions have no detectable side-effects."*

Source: [The recent released (at Google I/O 2018) style guide from the Android Team](https://android.github.io/kotlin-guides/style.html#constant-names).

philipbrito: From the same source of reference:

*"Source files are usually read from top-to-bottom meaning that the order, in general, should reflect that the declarations higher up will inform understanding of those farther down. Different files may choose to order their contents differently. Similarly, one file may contain 100 properties, another 10 functions, and yet another a single class."*

philipbrito (approved):

Issue #857, opened 2018-03-06
[NEW] Allow the user to create a new channel


aniketsingh03:
Rocket.Chat.Android version: latest build from `develop-2.x` branch.

**Description**
Presently the app does not allow the user to create (neither public/nor private) channels in the app.

**How can it be done**
This can be accomplished by adding a `FAB` to the chat rooms list fragment and then opening a new fragment where the user is asked about channel info (name, when was it created, who created it), and whether it must be private or public. The design will be quite similar to the web app.    

Would you like to work on this one?
**Yes**

aniketsingh03:
@rafaelks @filipedelimabrito can I work on this one?

aniketsingh03:
@filipedelimabrito is this in progress already?

rafaelks:
@filipedelimabrito @aniketsingh03 It's not in progress yet, you can start working on it! Thank you!

PR #891, opened 2018-03-12
[NEW] Add functionality to allow user to create new chat rooms

aniketsingh03:
@RocketChat/android

Closes https://github.com/RocketChat/Rocket.Chat.Android/issues/857

This pr aims to complete the following features:
- [x] Add FAB to create new channel.
- [x] Add a view to create a new channel.
- [x] Allow creation of new public channels and private groups
- [x] Call the rest api function to successfully create a new channel 
- [x] Add a view to invite members to the new channel.
- [x] Send list of members along with channel name in the rest call.
- [x] Support infinite scrolling for search results.
- [x] Fix CI tests

PR #1355, opened 2018-06-11
[NEW] Create channel

philipbrito:
Closes https://github.com/RocketChat/Rocket.Chat.Android/issues/857 and https://github.com/RocketChat/Rocket.Chat.Android/pull/891

I'd like to say thanks to @aniketsingh03 for his (great) idea of using chips to show the selected member. 

In this PR I'm removing the usage of the Float Action Button (action is "called" from the nav. drawer now) and using a fragment instead of activities (users will be able to add the members from the same view that they create a channel (respecting the user experience) - also after create a channel successfully we'll show the chat list with the recent created channel).

 ~~ATTENTION: There are some bugs with our API that returns some users with null usernames and that does not work when using the dot (`.`) character in the query. For more info see https://github.com/RocketChat/Rocket.Chat/issues/11091  and https://github.com/RocketChat/Rocket.Chat/issues/11092~~ 
 
Depends on https://github.com/RocketChat/Rocket.Chat.Kotlin.SDK/pull/178

Preview:

screenshot_20180611-144713


sampaiodiego:
you guys should be using the `spotlight` endpoint to search users, just like the web client does, this way we'll have the same results on both clients

rafaelks:
@sampaiodiego I think we'll change to use the `directory` API, how about that? We're using on iOS already.

sampaiodiego:
@rafaelks I think using `directory` endpoint is fine. they're almost the same on the implementation.

philipbrito:
@rafaelks @sampaiodiego Ok, we have that flexibility. But I'll end up using the `spotlight` endpoint that is already present in our SDK.

Android is hugging the web client. But don't feel alone on the iOS side. I'm sorry @rafaelks 😁

philipbrito:
Oops, is the web client moving to `directory` too? Then I think that it is better to user that API, so the web client and all mobile client will be using the same, right?

Give me some advice 🙏

rafaelks:
@filipedelimabrito `spotlight` has limited resources (returns 5 results maximum) and I don't think we should be using for this situation.

philipbrito:
@rafaelks Got it! Thanks for your advice. 

But before this discussion about what API should we use, I was searching for the maximum of 5 users (See [here](https://github.com/RocketChat/Rocket.Chat.Android/blob/9ddd590f702fa6b9558871863d6b91b086406085/app/src/main/java/chat/rocket/android/createchannel/presentation/CreateChannelPresenter.kt#L57))

Because we are using a small view to show the results (see the image on this PR description), then I think that's fine in this case.

Do you agree? Let me know.

rafaelks:
@filipedelimabrito Yes, the way you're doing today on Android makes sense to keep on 5. 👍

sampaiodiego:
just to clarify, I wrongly thought the web client was using `spotlight` but it is not.. on the web we're using a DDP subscription called `userAutocomplete` .. 🤷‍♂️

app/src/main/java/chat/rocket/android/createchannel/di/CreateChannelModule.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import android.arch.lifecycle.LifecycleOwner
+import chat.rocket.android.createchannel.presentation.CreateChannelView
+import chat.rocket.android.createchannel.ui.CreateChannelFragment
+import chat.rocket.android.dagger.scope.PerFragment
+import dagger.Module
+import dagger.Provides
+
+@Module
+@PerFragment
```
luciofm: Scoping is per `@Provides` not per `@Module`

luciofm: I've made some changes on Dagger, take a look how it is now, we were reusing the same `Job` instance for various fragments

app/src/main/java/chat/rocket/android/createchannel/di/CreateChannelProvider.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import chat.rocket.android.createchannel.ui.CreateChannelFragment
+import dagger.Module
+import dagger.android.ContributesAndroidInjector
+
+@Module
+abstract class CreateChannelProvider {
+
+    @ContributesAndroidInjector(modules = [CreateChannelModule::class])
+    abstract fun provideCreateChannelFragment(): CreateChannelFragment
```
luciofm: `@PerFragment` for the fragment

app/src/main/java/chat/rocket/android/createchannel/presentation/CreateChannelPresenter.kt
```diff
@@ -0,0 +1,71 @@
...
+import chat.rocket.common.util.ifNull
+import chat.rocket.core.RocketChatClient
+import chat.rocket.core.internal.rest.createChannel
+import chat.rocket.core.internal.rest.spotlight
+import javax.inject.Inject
+
+class CreateChannelPresenter @Inject constructor(
+    private val view: CreateChannelView,
+    private val strategy: CancelStrategy,
+    private val mapper: MemberViewModelMapper,
```
luciofm: Since we started using architecture components `ViewModel`, maybe it would be better to rename our UI models to `UiModel` like `MemberUiModel`, `MemberUiModelMapper`

app/src/main/java/chat/rocket/android/createchannel/ui/CreateChannelFragment.kt
```diff
@@ -0,0 +1,324 @@
...
+        }
+    }
+
+    override fun showGenericErrorMessage() {
+        showMessage(getString(R.string.msg_generic_error))
+    }
+
+    override fun showUserSuggestion(dataSet: List<MemberUiModel>) {
+        // Hiding the progress because we are showing it already.
+        hideSuggestionViewInProgress()
```
luciofm: This should be handled on the presentation layer

app/src/main/java/chat/rocket/android/createchannel/presentation/CreateChannelPresenter.kt
```diff
@@ -0,0 +1,71 @@
...
+                view.enableUserInput()
+            }
+        }
+    }
+
+    fun searchUser(query: String) {
+        launchUI(strategy) {
+            try {
+                val users = client.spotlight(query).users
+                val memberViewModelMapper = mapper.mapToUiModelList(users)
```
luciofm: `memberViewModelMapper` => `memberUiModelMapper` 😛

app/src/main/java/chat/rocket/android/members/adapter/MembersAdapter.kt
```diff
@@ -10,15 +10,24 @@ import chat.rocket.android.util.extensions.inflate
...
-    override fun onBindViewHolder(holder: MembersAdapter.ViewHolder, position: Int) = holder.bind(dataSet[position], listener)
+    override fun onBindViewHolder(holder: MembersAdapter.ViewHolder, position: Int) =
+        holder.bind(dataSet[position], listener)
 
     override fun getItemCount(): Int = dataSet.size
 
+    fun clearData() {
+        val itemCount = dataSet.size
+        dataSet = emptyList()
+        notifyItemRangeRemoved(0, itemCount)
```
luciofm: If you are removing everything, a simple `notifyDataSetChanged()` is enough

app/src/main/java/chat/rocket/android/createchannel/presentation/CreateChannelPresenter.kt
philipbrito: 😛

app/src/main/java/chat/rocket/android/createchannel/ui/CreateChannelFragment.kt
```diff
@@ -0,0 +1,306 @@
...
+        }
+    }
+    private val compositeDisposable = CompositeDisposable()
+    private var channelType: String = RoomType.CHANNEL
+    private var isChannelReadOnly: Boolean = false
+    private var memberList = arrayListOf<String>()
+
+    companion object {
+        fun newInstance() = CreateChannelFragment()
+    }
```
leonardoaramaki, line 48: Shouldn't we put the companion object at the end of the class according to our guideline?

luciofm (approved):

Issue #282, opened 2017-05-10
[IMPROVEMENT] Compress pictures on upload


TheReal1604:
Please add a feature that compresses the pictures before uploading this to a rocket.chat channel.

PR #1384, opened 2018-06-15
[NEW] Compress image

philipbrito:
Closes https://github.com/RocketChat/Rocket.Chat.Android/issues/282


app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomPresenter.kt
```diff
@@ -265,19 +271,37 @@ class ChatRoomPresenter @Inject constructor(
...
-                            client.uploadFile(roomId, fileName!!, mimeType, msg, description = fileName) {
-                                uriInteractor.getInputStream(uri)
+                withContext(DefaultDispatcher) {
+                    val fileName = uriInteractor.getFileName(uri) ?: uri.toString()
+                    val fileSize = uriInteractor.getFileSize(uri)
+                    val mimeType = uriInteractor.getMimeType(uri)
+                    val maxFileSizeAllowed = settings.uploadMaxFileSize()
+
+                    when {
+                        fileName.isEmpty() -> {
```
luciofm, line 281: Any reason to not use the single line indentation here?

app/src/main/java/chat/rocket/android/util/extensions/Image.kt
```diff
@@ -0,0 +1,19 @@
+package chat.rocket.android.util.extensions
+
+import android.graphics.Bitmap
+
+fun Bitmap.getCompressFormat(mimeType: String): Bitmap.CompressFormat {
```
luciofm, line 5: This can be simplified as
```kotlin
fun Bitmap.getCompressFormat(mimeType: String): Bitmap.CompressFormat {
    return when {
        mimeType.contains("jpeg") -> Bitmap.CompressFormat.JPEG
        mimeType.contains("webp") -> Bitmap.CompressFormat.WEBP
        else -> Bitmap.CompressFormat.PNG
    }
}
```

app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomPresenter.kt
philipbrito: Yes. 

It is allowed.
It is on our code base as well.
It isn't an "avoid recommendation".
Like `if(something) in the same line` or `if(something) {in a new line}`

app/src/main/java/chat/rocket/android/util/extensions/Image.kt
philipbrito: Very good! Thanks for pointing 👍

app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomPresenter.kt
luciofm: Actually it is a "recommendation", but it was more about curiosity 😛

>Put short branches on the same line as the condition, without braces.
```kotlin
when (foo) {
    true -> bar() // good
    false -> { baz() } // bad
}
```

luciofm (approved):
LGTM 👍

thepowerprocess:
Is this an option or are all images being compressed when uploading from Android app?

PR #1393, opened 2018-06-19
[IMPROVEMENT] Confirm before deleting the message

### LINKED ISSUE #1376: [IMPROVEMENT] Confirm before deleting the message
## Description
I think its better to confirm before deleting the message(we can use alert dialog) as delete option can be pressed by mistake.

divyanshub024:
@RocketChat/android

Closes #1376 
device-2018-06-19-170849


philipbrito (approved):
Yes!
Thank you @divyanshub024

app/src/main/java/chat/rocket/android/chatroom/adapter/ChatRoomAdapter.kt
```diff
@@ -18,7 +20,8 @@ class ChatRoomAdapter(
     private val roomName: String? = null,
     private val presenter: ChatRoomPresenter? = null,
     private val enableActions: Boolean = true,
-    private val reactionListener: EmojiReactionListener? = null
+    private val reactionListener: EmojiReactionListener? = null,
+    private val activity: FragmentActivity? = null
```
philipbrito: What about declaring the `context: Context? = null` instead of  `activity: FragmentActivity?` here?
Could you change it?

divyanshub024: Done :+1:

PR #1446, opened 2018-07-06
[IMPROVEMENT] Added project improvements.

philipbrito:
- Better app modularization (core and util module added). 
- Processing the draw image in background.
- Handling possible errors while processing the draw image.
- Removes the draw image being saved on the user device every time that it is generated.
- Add extensions to ChatRoomFragment.kt by adding Attachments.kt


draw/src/main/java/chat/rocket/android/draw/main/ui/DrawActivity.kt
```diff
@@ -181,4 +220,4 @@ class DrawingActivity : AppCompatActivity() {
 
     private val Int.toPx: Float
         get() = (this * Resources.getSystem().displayMetrics.density)
```
leonardoaramaki, line 222: :+1:

draw/src/main/java/chat/rocket/android/draw/main/ui/DrawActivity.kt
```diff
@@ -1,128 +1,167 @@
-package chat.rocket.android.draw
+package chat.rocket.android.draw.main.ui
```
leonardoaramaki, line 1: why `.main`?

leonardoaramaki (approved):

leonardoaramaki:
Seems fine to me. Just solving some conflicts are ready to go :+1:

draw/src/main/java/chat/rocket/android/draw/main/ui/DrawActivity.kt
philipbrito: I was not too much creative to define that package name. I just left `.main` because it is the "place"  of the main classes (activity) in that module.

PR #1451, opened 2018-07-08
[FIX] Fix Reply with attachment

### LINKED ISSUE #1447: [BUG] Reply with attachments and the new drawing feature does not work
@RocketChat/android 

Reply to any message, attach an image or any other file or draw something and hit send.
The message with the attachment/drawing is sent as a normal message, not as a reply.

shubhsherl:
@RocketChat/android

Closes #1447 

This PR resolves the issue of file attachment in a reply.


philipbrito (approved):
Thanks for submitting this PR @shubhsherl

PR #1466, opened 2018-07-10
[FIX] Message retain

### LINKED ISSUE #1464: [BUG] Message should retain in message-composer
## Description
While replying a message, shouldn't the message be retained in message-composer if I cancel the message_action_bar(i.e. cancel the reply), as it happens in web browser..? 

## Devices and Versions

Your Rocket.Chat.Android version: 2.4.0

Mobile device model and OS version: Xiaomi Redmi Note 4 - Android 7.0

## Steps to reproduce
Reply a message and write in the message box, and then cancel reply

shubhsherl:
@RocketChat/android

Closes #1464 

Fix the message retain issue when there is text in message composer and user cancel the reply option.


app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
```diff
@@ -914,7 +915,8 @@ class ChatRoomFragment : Fragment(), ChatRoomView, EmojiKeyboardListener, EmojiR
     private fun setupActionSnackbar() {
         actionSnackbar = ActionSnackbar.make(message_list_container, parser = parser)
         actionSnackbar.cancelView.setOnClickListener {
-            clearMessageComposition()
+            clearMessageComposition(false)
+            if (text_message.textContent.isEmpty())
             KeyboardHelper.showSoftKeyboard(text_message)
```
philipbrito: Could you change it to:
```
if (text_message.textContent.isEmpty()) {
    KeyboardHelper.showSoftKeyboard(text_message)	             
}
``` 
?

philipbrito (changes requested):
Could you only change a small piece of the code to just follow the coding style?

Thanks.

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
shubhsherl: Okay..

philipbrito (approved):
Thanks for submitting this PR @shubhsherl!

PR #1485, opened 2018-07-13
[RELEASE] Merge develop into beta

rafaelks:
@RocketChat/android


leonardoaramaki:
@filipedelimabrito I think he means to set the IDE to warn about missing newlines at the end of files. 
That's because having a newline at the end of file is seen as a good practice, so that when poking around with files on console (grep'ing, etc) you end with a non-messy output.

PR #1496, opened 2018-07-16
[I18N] A few more spanish translations

ear-dev:
@RocketChat/android

I've taken a stab at most of the remaining spanish translations.  It might be good to have a native spanish speaker review, but I think it's pretty straight forward.  Thanks.


philipbrito (approved):
Thank you very much for removing a lot of *TODOs*!
Great @ear-dev!

PR #1548, opened 2018-07-26
[FIX] Fix uppercase url scheme

### LINKED ISSUE #1546: [BUG] Links with uppercase on the scheme are not working
Refer to: https://github.com/RocketChat/Rocket.Chat.Android/pull/967

luciofm:
closes #967 
closes #1546

### DIFF
--- a/app/src/main/java/chat/rocket/android/util/extensions/View.kt
+++ b/app/src/main/java/chat/rocket/android/util/extensions/View.kt
@@ -7,15 +7,27 @@ import android.view.View
 import chat.rocket.android.R
 import timber.log.Timber
 
-fun View.openTabbedUrl(url: Uri) {
+fun View.openTabbedUrl(url: String) {
     with(this) {
+        val uri = url.ensureScheme()
         val tabsbuilder = CustomTabsIntent.Builder()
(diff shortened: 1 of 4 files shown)


app/src/main/java/chat/rocket/android/util/extensions/View.kt
```diff
@@ -7,15 +7,27 @@ import android.view.View
...
         } catch (ex: Exception) {
             Timber.d(ex, "Unable to launch URL")
         }
     }
+}
+
+private fun String.ensureScheme(): Uri? {
+    // check if the URL starts with a http(s) scheme
+    val url = if (!this.matches(Regex("^([h|H][t|T][t|T][p|P]).*"))) {
+        "http://$this"
```
rafaelks, line 27: What if the scheme is HTTPS?

luciofm: This is just to check if the url already has a scheme... Sometimes we receive urls like "www.something.com", so we append "http://"

rafaelks: Gotcha! 👍

PR #1562, opened 2018-07-30
[NEW] Key icon

philipbrito:
@divyanshub024 Let me see if this icon works for you.


app/src/main/res/drawable/ic_key_16dp.xml
```diff
@@ -0,0 +1,20 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="16dp"
+    android:height="16dp"
```
divyanshub024: Can we make height and width to 20 dp?

divyanshub024 (changes requested):
Naming the file as "ic_key_black_16dp" will be more helpful.

philipbrito:
@divyanshub024 The icon isn't black, so I don't call it `ic_key_black_16dp`. Btw, the icons that you are adding are filled in black?

app/src/main/res/drawable/ic_key_16dp.xml
philipbrito: Since it is a vector drawable you can define any height and width directly from the `imageView`. 
Btw, Android stands those kind of icon to 24dp. What are the `heights ` and `widths` that are you defining for that new icons?

divyanshub024: @filipedelimabrito We can change the width and height in imageview but I'm using those icons in button drawable(In which it would we bit difficult to change width and height).

divyanshub024: In design the icon size is 20dp, that's why I think It's better to make it 20dp.

divyanshub024:
@filipedelimabrito I'm not using black color, the icon is same color as givien in design. I was thinking about this name(`ic_key_black_16dp`) because it helps to let you know about the color without opening it. But this isn't black color so we can leave the name as it was.

app/src/main/res/drawable/ic_key_16dp.xml
philipbrito: I'll change it 👍

philipbrito:
@divyanshub024 We shouldn't call it `..._black_...dp` since it isn't black.
I changed the icon, take a look and let me know if it works for you.

divyanshub024 (approved):
Looks good! Thank you :+1:

philipbrito:
Thank you too @divyanshub024 👍

Issue #1508, opened 2018-07-18
[NEW] Open Admin panel


rafaelks:
## Description

Open this URL in a webview:

> "{baseURL}/admin/info?layout=embedded"

Once the page is loaded, execute this JavaScript code, the `{token}` value needs to be substituted by the user authentication token:

>Meteor.loginWithToken('{token}', function() { })

After this JavaScript gets executed, the page needs to open the Administration page automatically.

PR #1570, opened 2018-08-01
[NEW] Adds the admin panel.

philipbrito:
Closes https://github.com/RocketChat/Rocket.Chat.Android/issues/1508

After merging this PR, please, merge this another one in order to make it work: https://github.com/RocketChat/Rocket.Chat.Android/pull/1582


app/src/main/java/chat/rocket/android/main/ui/Menu.kt
```diff
@@ -0,0 +1,78 @@
+package chat.rocket.android.main.ui
+
+import android.view.Menu
+import android.view.MenuItem
+import chat.rocket.android.R
+
+private const val MENU_SECTION_ONE = 1
```
luciofm: It is better to keep `groupId`s and `itemId`s on `res/values/ids.xml` like: 
```
    <item name="menu_section_one" type="id"/>
    <item name="menu_action_chats" type="id"/>
```
This way we keep the ids unique, and don't risk to have id conflicts when adding more menu items on a Fragment for example.

Looking down, there is no need to add the items on code, we can add the admin panel to the XML and just have directly on `MainActivity`
```
private fun setupMenu(menu: Menu) {
    menu.findItem(R.id.menu_action_admin_panel).isVisible =  permissions.canSeeTheAdminChannel()
}
```

app/src/main/java/chat/rocket/android/helper/UserHelper.kt
```diff
@@ -12,57 +11,37 @@ import javax.inject.Inject
...
     /**
-     * Whether current [User] is admin on the current server.
+     * Return the display name for the given [user].
+     * If setting 'Use_Real_Name' is true then the real name will be given, otherwise the username
+     * without the '@' is yielded.
      */
-    fun isAdmin(): Boolean {
-        return user()?.roles?.find { it.equals("admin", ignoreCase = true) } != null
+    fun displayName(user: SimpleUser): String {
+        val displayName: String? = getCurrentServerInteractor.get()?.let {
```
luciofm: Can be simplified as
```
    return getCurrentServerInteractor.get()?.let {
        if (settingsRepository.get(it).useRealName()) {
            user.name
        } else {
            user.username
        }
    }.orEmpty()
```
No need for a temporary variable.

app/src/main/java/chat/rocket/android/helper/UserHelper.kt
```diff
@@ -12,57 +11,37 @@ import javax.inject.Inject
...
+            if (settingsRepository.get(it).useRealName()) {
+                user.name
```
luciofm: IIRC we use `user.name ?: user.username` on other places

app/src/main/java/chat/rocket/android/main/ui/Menu.kt
```diff
@@ -0,0 +1,78 @@
...
+
+const val MENU_ACTION_CHATS = 1
+private const val MENU_ACTION_CREATE_CHANNEL = 2
+const val MENU_ACTION_PROFILE = 3
+private const val MENU_ACTION_SETTINGS = 4
+private const val MENU_ACTION_ADMIN_PANEL = 5
+private const val MENU_ACTION_LOGOUT = 6
+
+internal fun MainActivity.setupMenu(menu: Menu) {
+    menu.add(
```
luciofm: `with(menu)` would save a lot of `menu.` 😀

app/src/main/res/menu/navigation.xml
```diff
@@ -1,44 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<menu xmlns:android="http://schemas.android.com/apk/res/android">
```
luciofm, line 2: No need to move from XML to code, you can add the Admin panel to XML and just change the visibility base on permissions.
`menuItem.isVisible = false`

app/src/main/java/chat/rocket/android/server/domain/PermissionsInteractor.kt
```diff
@@ -71,6 +68,28 @@ class PermissionsInteractor @Inject constructor(
         } == true || userHelper.isAdmin()
     }
 
+
+    fun canSeeTheAdminChannel(): Boolean {
```
luciofm: I think there is a type here, wouldn't it be `canSeeTheAdminPanel()`?

luciofm (changes requested):
I think it is better to keep the menu on XML, or at least its items ids.
I've been bitten by that before, a lot of time debugging to notice that I as setting up menu items on both the Activity and on Fragments, and they ended up sharing ids because I was declaring them on code, just like here.

app/src/main/java/chat/rocket/android/server/domain/PermissionsInteractor.kt
philipbrito: Good catch! 
https://github.com/RocketChat/Rocket.Chat.Android/pull/1570/commits/07dfbcdc586441b574fb1bd59bcb11f35f129513

app/src/main/res/menu/navigation.xml
philipbrito: Really don't need to move from XML to code here, but I came across a specific situation where I needed to customize a search view programmatically (take a look on `Menu.kt` for the ChatRoom), so I find interesting declaring it here in the same way.

We can change it later if really needed.

app/src/main/java/chat/rocket/android/helper/UserHelper.kt
philipbrito: https://github.com/RocketChat/Rocket.Chat.Android/pull/1570/commits/95ea9692ec0688e87a9ad44a946ad0d7b0a9f7e7

app/src/main/java/chat/rocket/android/main/ui/Menu.kt
philipbrito: Not only for the risks by declaring the ids in the current way, but its more interesting putting them in the `ids.xml`.

Added! 
https://github.com/RocketChat/Rocket.Chat.Android/pull/1570/commits/008f9df4bc22124ae63da5d4397cbf07dce42620

app/src/main/java/chat/rocket/android/main/ui/Menu.kt
philipbrito: Good usage of with as discussed internally.
Changed!
e190306

app/src/main/java/chat/rocket/android/main/ui/Menu.kt
```diff
@@ -0,0 +1,69 @@
+package chat.rocket.android.main.ui
+
+import android.view.Menu
+import android.view.MenuItem
+import chat.rocket.android.R
+
+internal fun MainActivity.setupMenu(menu: Menu) {
```
luciofm, line 7: If this is used only on MainActivity, it could be declared directly there, no gain to be an extension method declared somewhere else

app/src/main/res/menu/navigation.xml
luciofm: Still think it is a lot of code that doesn't need to be code... Even for the ChatRoom, the only thing that really needs to be code is `stylizeSearchView()` and the listeners...

luciofm (approved):

luciofm (changes requested):
```
 /home/circleci/project/app/src/main/res/layout/fragment_admin_panel_web_view.xml:9: Error: This view is not constrained. It only has designtime positions, so it will jump to (0,0) at runtime unless you add the constraints [MissingConstraints]
      <WebView
       ~~~~~~~
```

app/src/main/java/chat/rocket/android/main/ui/Menu.kt
philipbrito: Avoiding a long file and taking advantage of decorator pattern by Kotlin native support.

philipbrito:
Interesting, the lack of constraints wasn't an error in the previous AS version, the most current version shows a warn to put them (👍).

Done.

Issue #1455, opened 2018-07-09
[NEW] Improve Analytics tracking


rafaelks:
## Description

- [x] Make an option in "Preferences" to enable/disable the Analytics tracking. It's by default `true`.
- [x] Integrate with Google Analytics for Firebase (GA)
- [x] Make sure Answers (from Fabric) is integrated/installed
- [x] Call the following events:
  - [x] Login
    - [x] For Answer (Answers has a default method for that)
    - [x] For GA (you can use `login`)
  - [x] Signup
    - [x] For Answer (Answers has a default method for that)
    - [x] For GA (you can use `sign_up `)
  - [x] `screen_view`, params: [`screen`: screen name (can be a generic class name)]
    - [x] For Answer
    - [x] For GA
  - [x] `message_sent`, params: [`subscription_type`: just the type of the subscription (`c`, `d` or `p` ), `server`: just the host part]
    - [x] For Answer
    - [x] For GA
  - [x] `media_upload`, params: [`subscription_type`: just the type of the subscription (`c`, `d` or `p` ), `media_type`: mimetype of the file]
    - [x] For Answer
    - [x] For GA
  - [x] `reaction`, params: [`subscription_type`: just the type of the subscription (`c`, `d` or `p` )]
    - [x] For Answer
    - [x] For GA
  - [x] `server_switch`, params: [`server_url`, `server_count`]
    - [x] For Answer
    - [x] For GA

PR #1595, opened 2018-08-10
[NEW] Analytics for Answer (Fabric).

philipbrito:
Closes partly #1455


app/build.gradle
```diff
@@ -112,6 +110,12 @@ dependencies {
     kapt libraries.daggerProcessor
     kapt libraries.daggerAndroidApt
 
+    implementation libraries.fcm
+    implementation libraries.firebaseAnalytics
+    implementation libraries.playServicesAuth
```
luciofm: `fcm`, `firebaseAnalytics` and `playServicesAuth` must be `playImplementation` only

app/src/main/java/chat/rocket/android/about/ui/AboutFragment.kt
```diff
@@ -4,29 +4,36 @@ import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
-import androidx.core.view.isVisible
 import androidx.fragment.app.Fragment
 import chat.rocket.android.BuildConfig
 import chat.rocket.android.R
 import chat.rocket.android.main.ui.MainActivity
+import chat.rocket.android.server.domain.AnalyticsTrackingInteractor
```
luciofm, line 11: I think we should abstract this a litle more `AnalyticsTrackingInteractor`, `AnalyticsTrackingRepository` and `AnswersEvent`. I'll write a bigger comment on it.

app/src/main/java/chat/rocket/android/authentication/login/presentation/LoginPresenter.kt
```diff
@@ -5,30 +5,8 @@ import chat.rocket.android.authentication.presentation.AuthenticationNavigator
...
-import chat.rocket.android.server.domain.isGitlabAuthenticationEnabled
-import chat.rocket.android.server.domain.isGoogleAuthenticationEnabled
-import chat.rocket.android.server.domain.isLdapAuthenticationEnabled
-import chat.rocket.android.server.domain.isLinkedinAuthenticationEnabled
-import chat.rocket.android.server.domain.isLoginFormEnabled
-import chat.rocket.android.server.domain.isWordpressAuthenticationEnabled
-import chat.rocket.android.server.domain.isPasswordResetEnabled
-import chat.rocket.android.server.domain.isRegistrationEnabledForNewUsers
-import chat.rocket.android.server.domain.wordpressUrl
+import chat.rocket.android.server.domain.*
```
luciofm, line 8: I think you need to change your configuration to avoid `.*` imports

app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomPresenter.kt
```diff
@@ -17,22 +17,13 @@ import chat.rocket.android.db.DatabaseManager
...
-import chat.rocket.android.server.domain.JobSchedulerInteractor
-import chat.rocket.android.server.domain.MessagesRepository
-import chat.rocket.android.server.domain.PermissionsInteractor
-import chat.rocket.android.server.domain.PublicSettings
-import chat.rocket.android.server.domain.RoomRepository
-import chat.rocket.android.server.domain.UsersRepository
-import chat.rocket.android.server.domain.uploadMaxFileSize
-import chat.rocket.android.server.domain.uploadMimeTypeFilter
-import chat.rocket.android.server.domain.useRealName
+import chat.rocket.android.server.domain.*
```
luciofm: another wildcard

app/src/main/java/chat/rocket/android/dagger/module/AppModule.kt
```diff
@@ -21,32 +21,8 @@ import chat.rocket.android.infrastructure.LocalRepository
...
-import chat.rocket.android.server.infraestructure.MemoryChatRoomsRepository
-import chat.rocket.android.server.infraestructure.MemoryRoomRepository
-import chat.rocket.android.server.infraestructure.MemoryUsersRepository
-import chat.rocket.android.server.infraestructure.SharedPreferencesAccountsRepository
-import chat.rocket.android.server.infraestructure.SharedPreferencesMessagesRepository
-import chat.rocket.android.server.infraestructure.SharedPreferencesPermissionsRepository
-import chat.rocket.android.server.infraestructure.SharedPreferencesSettingsRepository
-import chat.rocket.android.server.infraestructure.SharedPrefsConnectingServerRepository
-import chat.rocket.android.server.infraestructure.SharedPrefsCurrentServerRepository
+import chat.rocket.android.server.domain.*
```
luciofm: one more

app/src/main/java/chat/rocket/android/authentication/signup/ui/SignupFragment.kt
```diff
@@ -154,9 +159,7 @@ class SignupFragment : Fragment(), SignupView {
     }
 
     override fun saveSmartLockCredentials(id: String, password: String) {
-        activity?.let {
-            it.saveCredentials(id, password)
-        }
+        activity?.saveCredentials(id, password)
```
luciofm, line 168: 👓 👍

luciofm (changes requested):
About how to approach the tracking on the code. I'd would go with something like this (pseudo kotlin code):
First an interface for our tracking manager with all our events (declared on a single place)
```
interface TrackingManager {
    sealed class Event {
        object LoginWithUsername : Event()
        object LoginWithDeeplink : Event()
    }

    sealed class Screen {
        object About : Screen()
        object Login : Screen()
        object ChatRoom : Screen()
    }

    fun trackEvent(event: Event, vararg params: Pair<String, Any>)

    fun trackScreen(screen: Screen)
}
```
Then we would inject a real implementation with dagger. For the FOSS flavour, it would be a simple noop manager:
```
// FOSS build
class FossTrackingManager : TrackingManager {
    override fun trackEvent(event: TrackingManager.Event, vararg params: Pair<String, Any>) {
    }

    override fun trackScreen(screen: TrackingManager.Screen) {
    }
}
```

For the Play store build, we would need another abstraction, like a `Tracker` interface`
```
interface Tracker {
    fun trackEvent(event: String, vararg params: Pair<String, String>)
    fun trackScreen(screen: String)
}
```

And one implementation for each Analytics we use:
```
class AnswersTracker(tracker: Answers) : Tracker {
    override fun trackEvent(event: String, vararg params: Pair<String, String>) {
        TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
    }

    override fun trackScreen(screen: String) {
        TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
    }
}

class FirebaseTracker(tracker: FirebaseAnalytics) : Tracker {
    override fun trackEvent(event: String, vararg params: Pair<String, String>) {
        TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
    }

    override fun trackScreen(screen: String) {
        TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
    }
}
```

and our `PlayTrackingManager`:
```
// Play Build
class PlayTrackingManager(private val trackers: List<Tracker>) : TrackingManager {
    override fun trackEvent(event: TrackingManager.Event, vararg params: Pair<String, Any>) {
        if (tracking is not enabled) return

        trackers.forEach { tracker ->
            // Do some processing and track the event
            tracker.trackEvent(event, *params)
        }
    }

    override fun trackScreen(screen: TrackingManager.Screen) {
        if (tracking is not enabled) return

        trackers.forEach { tracker ->
            tracker.trackScreen(screen)
        }
    }
}
```

that way, we only need to do one simple call on each place, no `if`'s, no multiple tracking calls, etc.
`trackingManager.trackScreen(Screen.About)`

app/build.gradle
filipedelimabrito: Duplicated library declaration here. Thanks for the catch.

app/src/main/java/chat/rocket/android/authentication/login/presentation/LoginPresenter.kt
filipedelimabrito: 👁 👍

app/build.gradle
philipbrito: Duplicated library declaration here. Thanks for the catch.

app/src/main/java/chat/rocket/android/authentication/login/presentation/LoginPresenter.kt
philipbrito: Using Multiple IDE Versions. I forgot to configure one of them. 👁 👍 Thanks.

philipbrito:
@luciofm 

I liked the idea of using sealed classes but I don't like the idea of having `TrackingManager` and `Tracker` interfaces since they are closed equals (duplicates the same functions: `trackEvent` and `trackScreen`). We also don't need the usage of `varargs`as well as we can avoid the use of loops here (`trackers.forEach {...}`).

Instead I did the following: 

1 - Created a simple interface with the required events to log (`Analytics` interface).
2 - Create sealed classes related to events (`SubscriptionTypeEvent`, `ScreenViewEvent`, `AuthenticationEvent`). The pain is that every time we create a fragment we'll need to put it in the `ScreenViewEvent` class - we can start thinking on something better. Btw, we can take advantage of using the `ScreenViewEvent.screenName` instead of declaring `const` all the fragment. Example:

```
addFragment(ScreenViewEvent.MessageInfo.screenName, R.id.fragment_container) {
    newInstance(messageId = messageId)
}
```

3 - For now, our singleton class `AnalyticsManager` implements the `Analytics` interface and logs the Answers events, but in the future we can improve the design and support the Dagger instead of using lots of `ifs` for sure (to inject `analyticsTrackingInteractor` and `GetCurrentServerInteractor`).

Keeping it simple by removing unneeded things (while maintaining the abstraction) is also good at this point.

philipbrito:
Requested changes addressed (with some points to the used abstraction [here](https://github.com/RocketChat/Rocket.Chat.Android/pull/1595#issuecomment-415141401)). 
Merging this PR guys.

PR #1611, opened 2018-08-22
[FIX] Temporary workaround for HandlerContext error.

philipbrito:
Temporary workaround for HandlerContext error.

`Cannot access class 'Lkotlinx.coroutines.experimental.android.HandlerContext;'. Check your module classpath for missing or conflicting dependencies`


luciofm:
@filipedelimabrito don't merge, I have a simpler solution, will open a PR...

philipbrito:
Reopening since the proposed solution doesn't work from here. Related issue for tracking the error: https://youtrack.jetbrains.com/issue/KT-26224

PR #1656, opened 2018-09-06
[FIX] fix emoji crash

luciofm:
This is purely a workaround for launching 2.6.0, we definitely need to address it better. It is not guaranteed that it will always work, it is a race condition. Most of the time it won't crash at least.


philipbrito (approved):

rafaelks:
Great branch name! 👍 😂

PR #1659, opened 2018-09-06
[FIX] Check file size after compressing the image

philipbrito:


rafaelks (approved):

app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomPresenter.kt
```diff
@@ -353,21 +354,16 @@ class ChatRoomPresenter @Inject constructor(
             try {
                 withContext(DefaultDispatcher) {
                     val fileName = uriInteractor.getFileName(uri) ?: uri.toString()
-                    val fileSize = uriInteractor.getFileSize(uri)
                     val mimeType = uriInteractor.getMimeType(uri)
+                    val byteArray = bitmap?.compressImageAndGetByteArray(mimeType)
```
luciofm, line 358: Ok for now, but I think we should save the image to a file, and use an inputstream... getting the whole bitmap as a byte array can cause OOM and memory fragmentation

luciofm (approved):

app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomPresenter.kt
philipbrito: Thanks for pointing! It worth to mention that the bitmap in this function is already "overridden" by "Glide" - what help us to avoid OOM.

PR #1669, opened 2018-09-11
[FEATURE] Initial message db support

luciofm:


app/build.gradle
```diff
@@ -152,6 +152,8 @@ dependencies {
 
     implementation libraries.livedataKtx
 
+    implementation 'com.google.code.findbugs:jsr305:3.0.2'
```
leonardoaramaki, line 155: Shouldn't this go to dependencies.gradle?

luciofm: Yes, it should 👍

util/src/main/java/chat/rocket/android/util/extension/Boolean.kt
```diff
@@ -0,0 +1,3 @@
+package chat.rocket.android.util.extension
+
+fun Boolean?.orFalse(): Boolean = this ?: false
```
leonardoaramaki, line 3: :+1:

app/src/main/java/chat/rocket/android/chatroom/uimodel/suggestion/ChatRoomSuggestionUiModel.kt
```diff
@@ -5,5 +5,4 @@ import chat.rocket.android.suggestions.model.SuggestionModel
 class ChatRoomSuggestionUiModel(text: String,
```
philipbrito, line 5: Could you fix the indentation here?

app/src/main/java/chat/rocket/android/db/model/Messages.kt
```diff
@@ -0,0 +1,106 @@
...
+    var attachments: List<AttachmentEntity>? = null
+    @Relation(parentColumn = "id", entityColumn = "messageId")
+    var reactions: List<ReactionEntity>? = null
+    @Relation(parentColumn = "id", entityColumn = "messageId")
+    var channels: List<MessageChannels>? = null
+
+    override fun toString(): String {
+        return "PartialMessage(message=$message, senderName=$senderName, senderUsername=$senderUsername, editName=$editName, editUsername=$editUsername, urls=$urls, attachments=$attachments, reactions=$reactions, channels=$channels)"
+    }
+
```
philipbrito, line 98: Unneeded blank line.
This fails the (future) lint (if we have plans to have it here).

app/src/main/res/layout/item_color_attachment.xml
```diff
@@ -35,11 +36,24 @@
...
         tools:text="#5571 - User profile from SSO must not have password change option" />
 
+    <TextView
+        android:id="@+id/text_fields"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:visibility="gone"
+        android:layout_marginStart="8dp"
+        android:layout_marginTop="4dp"
+        app:layout_constraintTop_toBottomOf="@id/attachment_text"
```
philipbrito, line 46: There are some questions related on using `@id` instead of `@+id`. In the end prefer to use `@+id` to don't care about it (widget disposition, etc.) - but looks like it doesn't have influence with the Constraint Layout.

philipbrito (changes requested):

leonardoaramaki (approved):

rafaelks (approved):

app/src/main/res/layout/item_color_attachment.xml
luciofm: `@+id` should be use when "declaring" the id, for relations I prefer to use `@id` so it will fail if we remove the view with the declared id.

PR #1710, opened 2018-09-29
[I18N] Add turkish language support

ardabeyazoglu:
@RocketChat/android

I just added a Turkish language support.


ardabeyazoglu:
Looks like there is a memory issue in build-play-apk step.

leonardoaramaki (approved):

PR #1720, opened 2018-10-02
Add translations

shubhsherl:
@RocketChat/android

Add some translations


divyanshub024:
@shubhsherl May I know how do translated these strings.

shubhsherl:
@divyanshub024 I used Google translator and cross checked if the translation is correct by translating it to English again. Just to be double sure if the translation is correct.

divyanshub024:
@shubhsherl Google translator is not a reliable option to translate language even if you double check it. I suggest you to only translate the language you know.

shubhsherl:
Oh, okay...

rafaelks:
Thanks for the contribution @shubhsherl, but we prefer the translation from native/fluent speakers  only and not translations from Google Translate. If you are native speaker or fluent in any language you're more than welcome to help us with these languages. Thanks!

PR #1738, opened 2018-10-10
[IMPROVEMENT] Set members list toolbar title immediately

### LINKED ISSUE #1294: [IMPROVEMENT] In members list toolbar title is update after the member list is loaded
## Description

Should have similar feature like pinned message and favourite messages.

ezgif-5-b85e9d82ae

Tunous:
@RocketChat/android

Closes #1294

This pull request makes it so the toolbar title in Members List screen is updated immediately on launch. Before the change it would be updated only after loading total number of members.


rafaelks (approved):

app/src/main/res/values-es/strings.xml
```diff
@@ -9,7 +9,8 @@
     <string name="title_legal_terms">Términos legales</string>
     <string name="title_chats">Chats</string>
     <string name="title_profile">Perfil</string>
-    <string name="title_members">Miembros (%d)</string>
+    <string name="title_members">Miembros)</string>
```
luciofm, line 12: There is a typo here... an extra `)`

PR #1767, opened 2018-10-16
[FIX] 3.0.0-beta2 fixes

luciofm:


app/src/main/java/chat/rocket/android/helper/MessageParser.kt
```diff
@@ -223,7 +223,16 @@ class MessageParser @Inject constructor(
 
     class LinkVisitor(private val builder: SpannableBuilder) : AbstractVisitor() {
 
+        var parsed = false
+
         override fun visit(text: Text) {
+            // XXX - Gambiarra ahead.
```
rafaelks: 🙈

app/src/main/java/chat/rocket/android/helper/MessageParser.kt
```diff
@@ -240,6 +249,8 @@ class MessageParser @Inject constructor(
                     consumed.add(link)
                 }
             }
+
+            parsed = true
```
rafaelks: 👍

rafaelks: @luciofm If the message gets edited, will it parse again?

PR #1853, opened 2018-11-13
[NEW] Add FAB, Show list of contacts, Invite Contacts via Email / Phone

kb0304:
@RocketChat/android

Closes https://github.com/WideChat/Rocket.Chat.Android/issues/124 , https://github.com/WideChat/Rocket.Chat.Android/issues/123

Demo : 
https://drive.google.com/file/d/14FeWFq6UTF7mxVcNoF2NLXJ8lUBtC-Th/view?usp=sharing

Depends upon: 
https://github.com/RocketChat/Rocket.Chat.Kotlin.SDK/pull/218
https://github.com/RocketChat/Rocket.Chat/pull/12617


rafaelks:
@kb0304 You will need to use the same branch name on the SDK to run the tests here.

kb0304:
@rafaelks I have created a PR to the SDK, could you please create a new branch in the SDK?

philipbrito (commented):
This is an interesting feature. My only concern is about the usage of the twilio API since it involves payments (https://www.twilio.com/pricing).

maybe @rafaelks can put his thoughts here.

bizzbyster:
@filipedelimabrito the plan is to create a configuration setting which specifies which custom field in the profile to use to store mobile phone number. If that is not set, then mobile phone number is not part of the profile and invites/contacts will only be supported via email. Of course, if an admin has Twilio setup then they can take advantage of the mobile phone number aspect of the feature as well. Does that make sense to you and @rafaelks ?

rafaelks:
@bizzbyster Yes, I guess we can use this setting, but why do we need Twilio on the mobile-side? To send the invitation to the user? Shouldn't this call be made from the back-end?

PR #1872, opened 2018-11-25
[NEW] User details screen

### LINKED ISSUE #1719: [NEW] User information screen
## Description

In the desktop app and in the browser, it's possible to display the information about a user. In the Android App, clicking on the user's name or image opens the dialog to respond to a message.

Ultimately, it is more convenient to get the message dialog by clicking on the message than using the "option" at the line end of the message. But the Android app should also allow you to view the information about a user.

## Devices and Versions

Your Rocket.Chat.Android version: 2.6.1
Your Rocket.Chat Server version: 0.69.2

## Steps to reproduce

Click on the user's name or image

leonardoaramaki:
@RocketChat/android

Closes #1719


app/src/main/java/chat/rocket/android/chatroom/adapter/MessageViewHolder.kt
```diff
@@ -70,9 +72,24 @@ class MessageViewHolder(
                 )
                 read_receipt_view.isVisible = true
             }
+
+            val senderId = data.message.sender?.id
+            val subscriptionId = data.subscriptionId
+            text_sender.setOnClickListener {
```
philipbrito, line 78: Add a space for better code readability like you did bellow.

app/src/main/java/chat/rocket/android/util/extensions/Graphics.kt
```diff
@@ -0,0 +1,25 @@
+import android.content.Context
+import android.graphics.Bitmap
+import android.renderscript.Allocation
+import android.renderscript.Element
+import android.renderscript.RenderScript
+import android.renderscript.ScriptIntrinsicBlur
+
+
```
philipbrito, line 8: You can add `@Throws(IllegalStateException::class)` here.

app/src/main/res/values-de/strings.xml
```diff
@@ -320,5 +320,12 @@
     <string name="notif_success_sending">Nachricht gesendet nach %1$s!</string>
     <string name="read_by">Gelesen von</string>
     <string name="message_information_title">Nachricht Information</string>
-    <string name="message_room_changed_privacy">Room type changed to: %1$s by %2$s</string>     <!--TODO - Add proper translation-->
+    <!--TODO - Add proper translation-->
```
philipbrito: Try to put the `<!-- TODO Add proper translation-->` in the side of the string so we can follow the same "pattern" that we're doing for all the TODOS in this file (as well as for the other `strings.xml` files too)

P.s: `<!--TODO - Add proper translation-->` instead of `<!--TODO - Add proper translation-->`

philipbrito (approved):
Not sure, but maybe we can put the analytics when the user opens the user details view here? 🤔 

@rafaelks Do you think this would be good?

rafaelks:
@filipedelimabrito @leonardoaramaki Yes, good finding! Let's add an event called: `user_details`.

rafaelks (changes requested):
Looks very nice so far, just few adjustments:

1. Missing custom fields support (you're on it AFAIK);
2. The image on the background is not filling correctly with the aspect ratio; Let's use "Aspect Fill" content mode;
3. Are you using `4dp` on the round corners?
4. When viewing the list of members, pressing in a user will open a different view. We can get rid of that view and use only the new one.

app/src/main/res/values-de/strings.xml
leonardoaramaki: Thought it was better to put above the line since we could have a longer string and thus that comment would be missed out. That's fine though... Putting them at the side

rafaelks (approved):

FaizanZahid:
nice :)

PR #1941, opened 2018-12-17
[i18N] Update strings.xml

TehPeGaSuS:
Fixed some syntax typos, that, imho, didn't make sense in the way that they were translated

Also translate some strings.

### DIFF
--- a/app/src/main/res/values-pt-rBR/strings.xml
+++ b/app/src/main/res/values-pt-rBR/strings.xml
@@ -221,8 +221,8 @@
     <string name="action_msg_delete">Remover</string>
     <string name="action_msg_pin">Pinar mensagem</string>
     <string name="action_msg_unpin">Despinar mensagem</string>
-    <string name="action_msg_star">Favoritar mensagem</string>
-    <string name="action_msg_unstar">Desfavoritar messagem</string>
+    <string name="action_msg_star">Marcar como favorita</string>
+    <string name="action_msg_unstar">Desmarcar como favorita</string>
     <string name="action_msg_share">Compartilhar</string>
     <string name="action_title_editing">Editando mensagem</string>
     <string name="action_msg_add_reaction">Adicionar reação</string>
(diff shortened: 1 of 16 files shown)


app/src/main/res/values-pt-rBR/strings.xml
```diff
@@ -219,10 +219,10 @@
     <string name="action_msg_copy">Copiar</string>
     <string name="action_msg_quote">Citar</string>
     <string name="action_msg_delete">Remover</string>
-    <string name="action_msg_pin">Pinar mensagem</string>
-    <string name="action_msg_unpin">Despinar mensagem</string>
-    <string name="action_msg_star">Favoritar mensagem</string>
-    <string name="action_msg_unstar">Desfavoritar messagem</string>
+    <string name="action_msg_pin">Afixar mensagem</string>
```
philipbrito: `Fixar mensagem` sounds better IMO. Both words exist in Brazilian Portuguese, but there is a little difference for them.

app/src/main/res/values-pt-rBR/strings.xml
```diff
@@ -219,10 +219,10 @@
...
+    <string name="action_msg_unpin">Desafixar mensagem</string>
```
philipbrito: `Desfixar` following the same idea above.

app/src/main/res/values-pt-rBR/strings.xml
philipbrito: https://github.com/RocketChat/Rocket.Chat.Android/pull/1941/commits/7259fe61092a0102ccf0c071e638b6b5053fb597

app/src/main/res/values-pt-rBR/strings.xml
philipbrito: https://github.com/RocketChat/Rocket.Chat.Android/pull/1941/commits/7259fe61092a0102ccf0c071e638b6b5053fb597

philipbrito (approved):

PR #1953, opened 2018-12-20
[FIX] Emoji keyboard

philipbrito:
This PR fixes:

- The bug that keeps the emoji keyboard open when changing the screen while it is open.
- The bug that keeps the emoji keyboard always opened after sending a message with a emoji at the end.


rafaelks (approved):
Looking really good! 👍

PR #1968, opened 2018-12-27
[I18N] Update Brazilian language translations

TehPeGaSuS:
Translated a newly added string and fixed some string positions

@RocketChat/android

The translation aims to also fix the position/sentence building syntaxes.
Let me know if anything is missing

### DIFF
--- a/app/src/main/res/values-pt-rBR/strings.xml
+++ b/app/src/main/res/values-pt-rBR/strings.xml
@@ -208,8 +208,8 @@
     <string name="message_pinned">Fixou uma mensagem:</string>
     <string name="message_muted">Usuário %1$s foi silenciado por %2$s</string>
     <string name="message_unmuted">Usuário %1$s saiu do modo silenciado por %2$s</string>
-    <string name="message_role_add">%3$s deu estatuto de %2$s a %1$s</string>
-    <string name="message_role_removed">%3$s removeu o estatuto de %2$s de %1$s</string>
+    <string name="message_role_add">%1$s foi definido %2$s por %3$s</string>
+    <string name="message_role_removed">%1$s não é mais %2$s por %3$s</string>
     <string name="message_credentials_saved_successfully">Credenciais salvas com sucesso</string>
 
     <!-- Message actions -->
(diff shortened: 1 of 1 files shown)


app/src/main/res/values-pt-rBR/strings.xml
```diff
@@ -205,13 +205,12 @@
...
     <string name="message_removed">Mensagem removida</string>
-    <string name="message_pinned">Pinou uma mensagem:</string>
+    <string name="message_pinned">Fixou uma mensagem:</string>
     <string name="message_muted">Usuário %1$s foi silenciado por %2$s</string>
     <string name="message_unmuted">Usuário %1$s saiu do modo silenciado por %2$s</string>
-    <string name="message_role_add">%1$s foi definido %2$s por %3$s</string>
-    <string name="message_role_removed">%1$s não é mais %2$s por %3$s</string>
-    // TODO:Add proper translation.
-    <string name="message_credentials_saved_successfully">Credentials saved successfully</string>
+    <string name="message_role_add">%3$s colocou o estatuto de %2$s a %1$s</string>
```
philipbrito: Why did you invert the order of the params here?

philipbrito (changes requested):
Are you translating it to Brazilian Portuguese or European Portuguese?

app/src/main/res/values-pt-rBR/strings.xml
TehPeGaSuS: I think it makes more sense something like:
@TheMythPT colocou o estatuto de administrador a @filipedelimabrito 
Probably changing `colocou` by `deu`?

TehPeGaSuS: When I get home, I'll make a few reviews.

TehPeGaSuS:
It would _**probably**_ be better to have a pt_PT language file?

philipbrito:
From what I could see it has been translated to European Portuguese from you - Am I right? If so we should definitely add a European Portuguese file to the app, because there are many differences to both language (pt_BR x pt_PT). Could you add that file?

Thank you 🇧🇷 🇵🇹

philipbrito (approved):
Thanks for this PR @TheMythPT! I'd removed some translations made for you here. Feel free to open a PR with a European Portuguese translation file as told above.

PR #1972, opened 2018-12-27
[CHORE] Update README.md

### LINKED ISSUE #1970: [CHORE] Update doc to tell users to use java 8 as java 11 does not work well with Android studio.
## Describe the chore

There are lots of issues saying that they cannot run rocket.chat app and it's not building. The main reason for this is due to having java 11(or above java 8) as your default java in your system. The android app project and Kotlin sdk project will not link and build unless java 8 is used.

divyanshub024:
- Specify Java version for project
- Update `Coding Style` section
- Removes the need of Android Studio Preview IDE to work with this project
- Uppercase some words

Closes #1970


philipbrito (approved):
Thank you @divyanshub024!

PR #1985, opened 2019-01-06
[I18N] Add Persian language

maryambehzi:
@RocketChat/android

#### Changes: 

Adds Persian language to the app


filipedelimabrito:
@maryambehzi I'd added TODOs for missing translations on you PR (more precisely to the Persian language file added from you). Could you please translate those strings?

I have the following concern: Will be the placeholders work from that file? Take a look on this example:

From the default (English) string file we have 
`<string name="message_role_add">%1$s was set %2$s by %3$s</string>`, from your file we have `<string name="message_role_add">%1$s به این تغییر داده شد %2$s توسط %3$s</string>`. Could you see the difference on how the placeholders (%1$s, %2$s, %3$s) are being used here?

philipbrito (commented):
@maryambehzi I've added TODOs for missing translations on you PR (more precisely to the Persian language file added from you). Could you please translate those strings?

I have the following concern: Will be the placeholders work from that file? Take a look on this example:

From the default (English) string file we have 
`<string name="message_role_add">%1$s was set %2$s by %3$s</string>`, from your file we have `<string name="message_role_add">%1$s به این تغییر داده شد %2$s توسط %3$s</string>`. Could you see the difference on how the placeholders (%1$s, %2$s, %3$s) are being used here?

philipbrito (approved):
@maryambehzi I've made changes on this PR to make sure it will work as expected. Basically all strings with placeholders (%1$d, %2$d, %d...) was updated - with a TODO wich indicates that those needs to be translated.

If you can translate those strings please open a PR with the translations making sure that the placeholders are declared as expect (as well as the \n escape).

Feel free to join our [channel](https://open.rocket.chat/channel/androidnativeapp) for any kind of question/support.

Thank you very much!

PR #1988, opened 2019-01-10
[FIX] Typing status

philipbrito:
Fix a bug that makes the typing status being send multiples time over the socket.


rafaelks (approved):
LGTM!

PR #2000, opened 2019-01-17
[I18N] Add translation for new italian strings

cmassi:
@RocketChat/android

Translated a couple of strings


philipbrito (approved):
Thank you very much @cmassi!

PR #2013, opened 2019-01-29
[REFACTOR] Coroutines yield

AllanWang:
@RocketChat/android

#### Changes: 

In looking through the code, I saw a recurrent use of `isActive`. Coroutines actually has a `yield()` function that will check for activity and throw a cancellation exception otherwise. Furthermore, `delay()` is a cancellable coroutine, so it will cancel on its own before and after the suspension.

Cancellation exceptions are caught within the live data by default with launch executions.

References:

Yield: https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/yield.html
Exception handling: https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-exception-handler/index.html


philipbrito (approved):
Thank you very much for sending this PR @AllanWang!
Also thanks for the well good explanation and the references!
💯

philipbrito:
@AllanWang I had to revert your PR here because it was breaking the CI. Do you care sending another PR making sure everything is working as expected?
Thanks.

AllanWang:
@filipedelimabrito are you sure the error is related to this PR? The last three CI errors all have different errors, none of which relating to the rocket chat package. If you see the relevant stacktrace, let me know, but otherwise I'd say try running the CI again

philipbrito:
@AllanWang The CI became failing after merging this PR. It was solved after reverting this PR. 
Take a look at this opened issue: https://github.com/RocketChat/Rocket.Chat.Android/issues/2119

Let's make a try again. Can you send a new PR with the changes made here?

philipbrito:
@AllanWang Do you want to submit your PR again?

AllanWang:
@filipedelimabrito I don't have an idea on what the problem is since it doesn't seem like a test failure, and if the error can't be replicated locally it's going to be pretty difficult for me to debug. If you want you can try to add the changes incrementally, but I can't tell how something like this will introduce a build error. If it's actually a test error it would help to know the class at fault

philipbrito:
We'll try again in a new PR. 
Thank you.
https://github.com/RocketChat/Rocket.Chat.Android/issues/2140

PR #2068, opened 2019-02-28
Fix about fragment title on rotate

### LINKED ISSUE #2063: [BUG] About fragment app bar title is change from "About" to "Rocket.Chat" on rotate.

arthurbdiniz:
Signed-off-by: Arthur Diniz <arthurbdiniz@gmail.com>

@RocketChat/android

Closes #2063

### DIFF
--- a/app/src/main/java/chat/rocket/android/about/ui/AboutFragment.kt
+++ b/app/src/main/java/chat/rocket/android/about/ui/AboutFragment.kt
@@ -40,6 +40,11 @@ class AboutFragment : Fragment() {
         analyticsManager.logScreenView(ScreenViewEvent.About)
     }
 
+    override fun onResume() {
+        super.onResume()
+        setupToolbar()
+    }
+
     private fun setupViews() {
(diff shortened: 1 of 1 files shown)


dev-ritik:
@arthurbdiniz this is not the correct way to handle this.
As mentioned [here](https://developer.android.com/guide/topics/resources/runtime-changes),
`Caution: Handling the configuration change yourself can make it much more difficult to use alternative resources, because the system does not automatically apply them for you. This technique should be considered a last resort when you must avoid restarts due to a configuration change and is not recommended for most applications.`

arthurbdiniz:
Hi @dev-ritik, thanks for the reply, i will fix this.

arthurbdiniz:
Just update the code with a more common solution

philipbrito:
@arthurbdiniz, thanks for sending this PR but this one sent one day ago before your change already fixes it: https://github.com/RocketChat/Rocket.Chat.Android/pull/2092

PR #2072, opened 2019-03-01
[BUG] Fix sending blank messages in reply and quote message.

### LINKED ISSUE #2070: [BUG] Should not send spaces in reply and quote message.

Cool-fire:
@RocketChat/android

Closes #2070


app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
```diff
@@ -868,8 +868,10 @@ class ChatRoomFragment : Fragment(), ChatRoomView, EmojiKeyboardListener, EmojiR
 
             button_send.setOnClickListener {
                 var textMessage = citation ?: ""
```
philipbrito: No need to have `var textMessage = citation ?: ""`
`if (!citation.isNullOrBlank()) { ... }` is the way to go.

philipbrito (changes requested):

app/src/main/java/chat/rocket/android/chatroom/ui/ChatRoomFragment.kt
Cool-fire: Thanks @filipedelimabrito for comments. I think it we should keep `var textMessage = citation ?: ""` we are adding `text_message.textContent` to textMessage. And textMessage holds **citation** if it is there else a empty string. For the bug I just added `if(!text_message.textContent.isBlank())` to check if the input message is blank and the line `var textMessage = citation ?: ""` exists from previous codebase .

Cool-fire: Let me know if I have to change the code. I will do it and update the pull request.

philipbrito (approved):
I've made some changes here.
Thanks for sending this PR @Cool-fire

PR #2131, opened 2019-03-13
Add native Licence Fragment

### LINKED ISSUE #2130: Native View of Showing Licence Screen. For Native UX
## Describe the feature you'd like
Instead of opening WebView of github's Licence Screen , native screen should be there for Licence screen

## Devices and Versions

Your Rocket.Chat.Android version: (e.g. 2.1.0)
Your Rocket.Chat Server version: (e.g. 0.63.1-develop)

RotBolt:
@RocketChat/android

Closes #2130 

#### Changes
Add Native License Fragment for showing license

### DIFF
--- a/app/src/main/java/chat/rocket/android/dagger/module/ActivityBuilder.kt
+++ b/app/src/main/java/chat/rocket/android/dagger/module/ActivityBuilder.kt
@@ -24,6 +24,7 @@ import chat.rocket.android.draw.main.di.DrawModule
 import chat.rocket.android.draw.main.ui.DrawingActivity
 import chat.rocket.android.favoritemessages.di.FavoriteMessagesFragmentProvider
 import chat.rocket.android.files.di.FilesFragmentProvider
+import chat.rocket.android.licence.di.LicenceFragmentProvider
 import chat.rocket.android.main.di.MainModule
 import chat.rocket.android.main.ui.MainActivity
 import chat.rocket.android.members.di.MembersFragmentProvider
@@ -67,6 +68,7 @@ abstract class ActivityBuilder {
             ProfileFragmentProvider::class,
             SettingsFragmentProvider::class,
             AboutFragmentProvider::class,
+            LicenceFragmentProvider::class,
             PreferencesFragmentProvider::class,
             AdminPanelWebViewFragmentProvider::class
         ]
--- a/app/src/main/java/chat/rocket/android/licence/di/LicenceFragmentProvider.kt
+++ b/app/src/main/java/chat/rocket/android/licence/di/LicenceFragmentProvider.kt
@@ -1,0 +1,12 @@
+package chat.rocket.android.licence.di
+
+import chat.rocket.android.licence.ui.LicenceFragment
+import dagger.Module
+import dagger.android.ContributesAndroidInjector
+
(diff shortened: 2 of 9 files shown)


philipbrito:
Thanks for sending this PR @theLimitBreaker but we really don't want to have it into the code. This is the reason: We'll need to update our Readme.md all the time the license here is updated and we also would need to update the application all the time to reflect that changes.

Opening a webview to show the license here in the Github (where the code is) is enough.

RotBolt:
> Thanks for sending this PR @theLimitBreaker but we really don't want to have it into the code. This is the reason: We'll need to update our Readme.md all the time the license here is updated and we also would need to update the application all the time to reflect that changes.
> 
> Opening a webview to show the license here in the Github (where the code is) is enough.

Hmm I understand, Previous Chat client I worked with  recommended  using native actvity/fragment. Thats why I thought using native View would be good. 
No Problem :)

TehPeGaSuS:
Dumb question: isn't there a way to `fetch` the license content, via the raw URL, of the file or in some other way and include it natively? I would be much more pleasant esthetically in the application.

Also not to mention that having things natively implemented are way better.

But that's just my 2 cents.

rafaelks:
Yes, totally: https://raw.githubusercontent.com/RocketChat/Rocket.Chat.Android/develop/LICENSE.

TehPeGaSuS:
and it's possible to `live fetch` that into the application?
If yes, would be something to consider.  :)

rafaelks:
Yes, that's totally possible (and the link is safe and will be able to show updates).

RotBolt:
Umm should I reopen this Issue , and this time will do a Live fetching of License ?

TehPeGaSuS:
I mean, WebView is handy but in this case would be better to display just the text in a native way than displaying the whole `front page` with the issues and stuff as it's currently displayed.

Hope that the idea is worthy of work 👍

TehPeGaSuS:
> Umm should I reopen this Issue , and this time will do a Live fetching of License ?

I believe, IMHO, that it deserves a second shot!
👍

TehPeGaSuS:
Also, it probably would fit better in the screen, since in my mobile (5" screen) I have to swipe right to see the totality of some lines.

RotBolt:
Glad ! Thank you @TheMythPT  , 
@rafaelks I have a question to ask , since its single licence fetching, Should I directly fetch the data directly with StreamReader in the fragment only ?

rafaelks:
@filipedelimabrito Would be the better to answer this question.

philipbrito:
I'm not sure if it's worth to do that inside the application. Honestly I don't see good reasons for doing that. A big number of well-known apps (Instagram, Twitter, Telegram, etc) use a webview to display its licence, privacy policy and terms of use for instance.

The best for this common use case is using a webview which shows the contents from a mobile friendly website (which is not the case of the Github license page as pointed above).

We don't need to complex the things adding more processing, etc, etc.

TehPeGaSuS:
Then, probably the best would be to replace the current URL by the raw URL

RotBolt:
Raw URL , does open in Mobile Friendly way ! . That's simple solution to do so

PR #2136, opened 2019-03-14
Adding clearMessageComposition function

### LINKED ISSUE #2135: [BUG] /me messages are sent to room and still kept in the input box
## Description

When sending `/me` messages via the android app, the text is sent to the room as `_message_` but its also kept in the input box after being sent.
On the same note, messages sent via `/me` are seen as `_message_` while those sent with `_message_` are properly displayed as _`message`_
## Devices and Versions

Your Rocket.Chat.Android version: 3.2.0
Your Rocket.Chat Server version: Official Rocket.chat server

Mobile device model and OS version: Huawei Y6 Pro 2017, Android 7.0

## Steps to reproduce
just type `/me message` and send it to the room. You'll get the described behaviour

AuraOfDivinity:
@RocketChat/android

Closes #2135


philipbrito (changes requested):
Thanks for clearing the message but it isn't closing the https://github.com/RocketChat/Rocket.Chat.Android/issues/2135 entirely as pointed above.

PR #2146, opened 2019-03-17
Fix the issue of message box not being cleared after sending the message

### LINKED ISSUE #2145: [BUG] Message box not being cleared for some time after sending a message
## Describe the bug
The message box is not cleared for some time after sending a message, particularly on slow connections.

## To Reproduce
Try sending messages on a slow connection. The bug cannot be reproduced if there is no connection.

## Screenshots
Screenshot_20190317-131704

## Devices and Versions

Your Rocket.Chat.Android version: Latest develop
Your Rocket.Chat Server version: open.rocket.chat

ubarsaiyan:
@RocketChat/android

Closes #2145


app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomPresenter.kt
```diff
@@ -352,6 +352,7 @@ class ChatRoomPresenter @Inject constructor(
                                 RoomUiModel(roles = chatRoles, isBroadcast = chatIsBroadcast)
                             ), false
                         )
+                        view.clearMessageComposition(true)
```
philipbrito, line 355: Doing that, if something went wrong when calling the ` client.sendMessage(...)` will cause the users lost their messages, needing it to be written again.

philipbrito (changes requested):

app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomPresenter.kt
ubarsaiyan: But the unsent message will still be displayed (greyed out) so the user won't lose the message. Am I wrong?

PR #2155, opened 2019-03-19
[i18n] Update DE strings.xml

AnBo83:
@RocketChat/android


philipbrito (approved):
Thank you very much for translating it @AnBo83!

PR #2209, opened 2019-04-08
[FIX] Migrate JitsiMeet api and fix build errors

water2015sit:
JitsiMeet changed their api recently, it caused our build errors. We
need to migrate our code to use their latest apis to fix build errors.

@RocketChat/android 
Hi, this is Runzhi, do you guys have time to review my code, it fixed the critical build error in our codebase.


philipbrito:
Thanks for sending this PR @water2015sit but we don't want remove the Jitsi functions we have for now but use an alternative instead.

PR #2228, opened 2019-04-16
[IMPROVEMENT] Hide the title of toolbar and change style of toolbar text

### LINKED ISSUE #2227: [BUG] In landscape view Toolbar shows server name and App name both
## Describe the bug
device-2019-04-16-153240

divyanshub024:
@RocketChat/android

Closes #2227


app/src/main/res/layout/app_bar_chat_rooms.xml
```diff
@@ -23,9 +23,7 @@
             android:drawablePadding="10dp"
             android:fontFamily="sans-serif-medium"
             android:maxLines="1"
-            android:textColor="#FFFFFF"
-            android:textSize="20sp"
-            android:textStyle="normal"
+            style="@style/TextAppearance.AppCompat.Widget.ActionBar.Title"
```
divyanshub024, line 26: @filipedelimabrito This is default style of the toolbar title. Having this style also helps us to have different style and font on the title on landscape and portrait mode. otherwise, in landscape, the title font will be the same(20sp) and will look quite big.

philipbrito: @divyanshub024 Thanks for sending this PR. The reason we've defined our own style is to follow our internal design specs. I'll take a look to see how it looks like here.

philipbrito:
@divyanshub024 I'm going to close this PR. We need to follow the design we have for the toolbar. Actually we'll show the server name and not its URL which will look better we for what have currently.
I'll close the #2227 from a new PR very soon (thanks for adding the `supportActionBar?.setDisplayShowTitleEnabled(false)` here).

The toolbar in landscape view:
Screen Shot 2019-04-17 at 15 37 02

PR #2251, opened 2019-04-21
[FIX] Can't install dev and release on the same device

Adizbek:
Was conflicting Fileprovider's authorities.

@RocketChat/android

#### Issue: 
Failed to commit install session 700417465 with command cmd package install-commit 700417465. Error: INSTALL_FAILED_CONFLICTING_PROVIDER: Package couldn't be installed in /data/app/chat.rocket.android.dev-NIX0violFlkc0WZmKcI6CQ==: Can't install because provider name chat.rocket.android.fileprovider (in package chat.rocket.android.dev) is already used by chat.rocket.android

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -115,7 +115,7 @@
 
         <provider
             android:name="androidx.core.content.FileProvider"
-            android:authorities="chat.rocket.android.fileprovider"
+            android:authorities=".fileprovider"
             android:exported="false"
             android:grantUriPermissions="true">
             <meta-data


philipbrito:
@Adizbek As the documentation states:

_"For your own app, specify an authority consisting of the app's android:package value with the string "fileprovider" appended to it."_
(https://developer.android.com/training/secure-file-sharing/setup-sharing#DefineProvider)

I know it can sound frustrating to you since the ability to use both build variantes is not possible. But do you know what such a change (declaring only `.fileprovider`) can lead for especially for other applications for instance?).

I'm going to close this PR, but if you want let us know why we should really include it. 
Thank you!

Adizbek:
@filipedelimabrito But it wouldn't affect to other apps, since release app's package will be: chat.rocket.android.fileprovider and for beta: chat.rocket.android.dev.fileprovider. They will act as different apps and will not conflict. And it will be handy to have both variants of app while developing and testing.

PR #2268, opened 2019-04-27
[IMPROVEMENT] Let language setter show the selected locale

hfeky:
@RocketChat/android

#### Changes:

Language setter now shows the current selected locale when the dialog is opened.

Language setter


app/src/main/java/chat/rocket/android/settings/ui/SettingsFragment.kt
```diff
@@ -163,29 +165,49 @@ class SettingsFragment : Fragment(), SettingsView, AppLanguageView {
 
     private fun changeLanguage() {
         context?.let {
+            val locales = ArrayList<Pair<String, String>>()
+            with(locales) {
+                add(Pair("en", ""))
```
philipbrito: Why to create a Pair if the country for some languages don't exist? It is useless in that case.

philipbrito (changes requested):

app/src/main/java/chat/rocket/android/settings/ui/SettingsFragment.kt
hfeky: I used Pair for all of them, since I just refer to it using `locale.first` and `locale.second` at the bottom. I can make it be an `arrayListOf<Any>()` and add String together with Pair but then it will make the implementation longer at the bottom, since I will need to check for the instance type if it is a Pair or just a String. Do you recommend using that approach I just said, or do you have any other suggestion? I am surely open to any efficient solution.

philipbrito: Why creating a Pair if the `second` is useless? Actually there is no need to create a Pair if we know that some languages don't have a country code. You are just wasting it and making everyone who works with that code creating an empty string in that cases

hfeky: I have fixed it. Please review the updated code. I added String for locales with language only, and added Pair for locales with language and country. I also changed the way how the current selected locale is read. Instead of reading it from the SharedPreferences, I get it from the current applied configuration. This is useful if language and country are unspecified; that is if the user never set a language from the settings before.

hfeky:
@filipedelimabrito Can you please review the changes done here?

app/src/main/java/chat/rocket/android/server/infrastructure/SharedPrefsCurrentLanguageRepository.kt
```diff
@@ -16,10 +17,10 @@ class SharedPrefsCurrentLanguageRepository(private val preferences: SharedPrefer
     }
 
     override fun getLanguage(): String? {
-        return preferences.getString(CURRENT_LANGUAGE, "")
+        return preferences.getString(CURRENT_LANGUAGE, Locale.getDefault().language)
```
philipbrito, line 20: Good one! Thanks!

philipbrito (approved):
Thanks for sending that PR @HusseinElFeky! 💯

app/src/main/java/chat/rocket/android/settings/ui/SettingsFragment.kt
```diff
@@ -163,28 +165,58 @@ class SettingsFragment : Fragment(), SettingsView, AppLanguageView {
 
     private fun changeLanguage() {
         context?.let {
+            // Add [String] for locales without specified countries, and
+            // add [Pair] for locales with specified countries.
+            val locales = arrayOf(
```
philipbrito: I think `locales` could be a constant...
We can make it a `const` later.

philipbrito (approved):

app/src/main/java/chat/rocket/android/settings/ui/SettingsFragment.kt
hfeky: Yes. That's a great idea. But how are we going to define a constant of non-primitives? Shall we convert every `Pair` to `String` separated by a delimiter? Like for example, `Pair("pt", "BR")` will be `"pt,BR"` and later split it below in the code? or is there a better way? And if yes, what is the best delimiter to use? A comma? Also shall we define it outside the class at the very top or in a `companion object`? or maybe create a `string-array` in a resource file? I'm still learning Kotlin and would like to know what the best practices are when we need to create a constant of non-primitives in Kotlin.

hfeky: I have refactored the code and added all the locales in a `string-array`. I feel it's neater that way. If you find the old approach better, just let me know, and I will undo the commit.

philipbrito: @HusseinElFeky You can put a constant on the top side of a class (look for some examples of it throughout our code as well as for discussions/recommendations on the web - there are good sources here).

Generally we make use of the strings.xml file(s) to things that will be displayed for the user with optional text styling and formatting - what is not the case.

The comma delimiter is ok.

app/src/main/res/values/strings.xml
```diff
@@ -105,6 +105,25 @@ https://github.com/RocketChat/java-code-styles/blob/master/CODING_STYLE.md#strin
         <item>Chinese (TW)</item>
     </string-array>
 
+    <string-array name="locales">
```
philipbrito: Let's keep it as a const as discussed above @HusseinElFeky.
Could you do that?

hfeky: I tried declaring the array as `private const val LOCALES = arrayOf(<locales>)` on the top side of the class, but unfortunately couldn't because `arrayOf(<elements>)` is not a primitive in Kotlin. :/ It has to be declared as `private val LOCALES = arrayOf(<locales>)`. Is that OK? or is there a workaround? And if that's OK, shall I keep all the array elements as `String` and use a comma as a delimiter, or revert back to using `String` with `Pair` together? Which is better?

philipbrito: @HusseinElFeky I've updated it*. Check it out.
Thank you very much for being active here and sharing with me interesting readings and being open-minded (just want to thank you again - now here). 👊 

*Since it is not a `const` there is no reason to keep it at the top the class.

hfeky: Yup, now I get it. You are very welcome! :fist_oncoming: Glad to help. :smiley:

PR #2277, opened 2019-05-01
[IMPROVEMENT] Chat sorting options Made visible in Landscape Mode

### LINKED ISSUE #2253: [BUG] Chat sorting options not visible in pop-up when phone orientation is landscape.
## Describe the bug

Chat sorting options not visible in pop-up when phone orientation is landscape.

## To Reproduce

1. Hold phone in horizontal orientation.
2. Tap the sorting button at the top right.
3. Sorting options are inaccessible.

## Expected behavior

Sorting options should be accessible in horizontal orientation.

## Devices and Versions

Your Rocket.Chat.Android version: 3.4.0

## GIF

issue horizontal-2253

Cool-fire:
@RocketChat/android

Closes #2253 

#### Changes:
 Added new style for bottomsheet which contains `behaviour_peekHeight` and also refactored `bottom_seet_fragment_directory_sorting` to `bottom_sheet_fragment_directory_sorting`

videotogif_2019 05 01_17 21 04


Cool-fire:
@filipedelimabrito Please review

app/src/main/res/values/styles.xml
```diff
@@ -269,4 +270,12 @@
         <item name="actionModeCloseDrawable">@drawable/ic_close_white_24dp</item>
     </style>
 
+    <style name="ChatRooms.BottomSheetDialog" parent="Theme.Design.Light.BottomSheetDialog">
+        <item name="bottomSheetStyle">@style/BottomSheetStyle</item>
+    </style>
+
+    <style name="BottomSheetStyle" parent="Widget.Design.BottomSheet.Modal">
+        <item name="behavior_peekHeight">512dp</item>
```
philipbrito, line 278: Where did you find this value? From an official reference?

Cool-fire: @filipedelimabrito I found it in bug tracker list of google and others had this problem so one of them suggested this solution. Is there anything wrong with this?

philipbrito: I was just asking. Can you send the bug tracker link for this? So I can (and others too) take a look on it too. Thank you.

Cool-fire: @filipedelimabrito here is the link https://issuetracker.google.com/issues/37132390#c6

philipbrito: @Cool-fire Thanks for providing us the link. Have you tested the `512dp` in more than one screen density?

I can't find references that it should be `512dp` from that source.

Cool-fire: @filipedelimabrito No I have tested on my phone and emulator seems its working fine.

HusseinElFeky: @Cool-fire Make sure it works as well on small-screen devices. I would also say it would be better if you can make the parent layout of every bottom sheet layout we have to have a height of `wrap_content` instead of `match_parent`, and add a ScrollView to every layout after the bottom sheet title divider, since at certain points, one may need to scroll if the list is very long on the device screen.

hfeky: @Cool-fire Make sure it works as well on small-screen devices. I would also say it would be better if you can make the parent layout of every bottom sheet layout we have to have a height of `wrap_content` instead of `match_parent`, and add a `ScrollView` to every layout after the bottom sheet title divider, since at certain points, one may need to scroll if the list is very long on the device screen.

philipbrito:
Closed by #2328

PR #2287, opened 2019-05-03
[IMPROVEMENT] Immediate message loading

RobotJINI:
@RocketChat/android

Closes #ISSUE_NUMBER

1.Changed the loading of moderator roles until after the room loads up so it does not block the loading of the ui. There isn't actually anything that uses the moderator roles right now, so i am leaving it to the implementer of that on how to best handle the delayed information. I wanted to make sure the functionality was the same after the changes though rather than just deleting it.

2. recycler_view was adding its adapter in showMessages this was causing the error: "E/RecyclerView: No adapter attached; skipping layout"
This seems to be the cause of a missed update step and ui delay. Code to add adapter was added to setupRecyclerView, which is called in the onViewCreated function.

3. loadActiveMembers function was being called from ChatRoomFragment object and the network / data access was being done in the UI thread. Moved the function call from the end of the function loadMessages to just after the end of said function. Changed the function loadActiveMembers to do most of its work in an I.O. thread and just the view updating in a UI thread.

roomLoadDemo


app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomPresenter.kt
```diff
@@ -135,47 +136,73 @@ class ChatRoomPresenter @Inject constructor(
...
+
+            chatRoomMessage?.let { messageHelper.messageIdFromPermalink(it) }
                     ?.let { messageId ->
                         val name = messageHelper.roomNameFromPermalink(chatRoomMessage)
                         citeMessage(
-                            name!!,
-                            messageHelper.roomTypeFromPermalink(chatRoomMessage)!!,
-                            messageId,
-                            true
+                                name!!,
```
philipbrito: @RobotJINI Don't forget to follow our Kotlin coding style (not sure if you're using another one on your repo/project).

app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomPresenter.kt
```diff
@@ -135,47 +136,73 @@ class ChatRoomPresenter @Inject constructor(
...
-                val room = dbManager.getRoom(roomId)
-                room?.let {
-                    chatIsBroadcast = it.chatRoom.broadcast ?: false
-                    val roomUiModel = roomMapper.map(it, true)
-                    launchUI(strategy) {
-                        view.onRoomUpdated(roomUiModel = roomUiModel.copy(
+            // Can post anyway if has the 'post-readonly' permission on server.
+            val room = dbManager.getRoom(roomId)
+            room?.let {
+                chatIsBroadcast = it.chatRoom.broadcast ?: false
```
philipbrito: `isSomething` generally fits better.

So:

`isChatBroadcast` or `isBroadcastChannel` instead of  `chatIsBroadcast`.

app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomPresenter.kt
```diff
@@ -135,47 +136,73 @@ class ChatRoomPresenter @Inject constructor(
...
+                    //view.updateModeration()
+                }
+            }
+
+            subscribeRoomChanges()
+        }
+    }
+
+    private suspend fun getChatRole() : Boolean {
+        var returnVal = false
```
philipbrito: What the variable `returnVal` stands for?

philipbrito (approved):
Thanks for sending this PR @RobotJINI!
I've updated the requested changes here.

PR #2289, opened 2019-05-04
[FIX] Fixed autosizing issue for notifications.

### LINKED ISSUE #2220: [BUG] Layout issues when device font is set to larger than default
## Describe the bug

When one sets the device font to be larger than default from the device settings, there are some layout issues that are produced in many parts of the app.

## Expected behavior

The app should optimize its user interface according to the device font set. This can easily be fixed by setting the views' dimensions to be in `sp` (Scale-independent Pixels) instead of any other unit, for any view that can change size when text size is changed.

## Screenshots

Some examples:

Screenshot

## Version

Rocket.Chat.Android version: 3.3.0

noor-binte-amir:
@RocketChat/android

Closes #2220 

#### Changes:
Added autosizing for TextView in xml.

## GIF:
-2220-layoutissue


app/src/main/res/layout/unread_messages_badge.xml
```diff
@@ -9,6 +9,7 @@
         style="@style/ChatList.Badge.TextView"
         android:layout_width="21dp"
         android:layout_height="21dp"
+        android:autoSizeTextType="uniform"
```
hfeky, line 12: I would say not to fix a font size to the text when device font size is changed. That autosize feature will still be limited to the view boundaries. It will be better to apply that font size change by changing the layout width and height units be in `sp` instead of `dp`. :slightly_smiling_face:

hfeky (changes requested):

app/src/main/res/layout/unread_messages_badge.xml
noor-binte-amir: Thankyou for the suggestion. Using `sp` is discouraged for width and height adjustments. I haven't come across a source that doesn't identify it as bad practice. Let me know if you have, I'd love to check it out as well!

hfeky: I couldn't come across a source that mentions setting views' width and height in `sp` unit is a bad practice. Why would it be? In certain cases like that, when changing device font size, all views' that had size units of `sp` will change dimensions. If we made the badge width and height be `wrap_content` and added a little padding to it, the badge would change size when the device font size is changed; however, since we need the badge to be squared, we set its dimensions instead of `wrap_content`, but rather the dimensions should be in `sp`, since it is a TextView and it will change size when device font size is changed just as `wrap_content` would have done. The `sp` unit is the same size as `dp`, by default, but it resizes based on the user's preferred text size.

noor-binte-amir: I was going with Android's developer's guide on screen densities which states that: "When defining text sizes, however, you should instead use scalable pixels (sp) as your units (but never use sp for layout sizes)."

However, you raise a good point. I've raised another PR with an alternative fix using `sp`. Curious to see what the solution is.

hfeky: Although that's true for most cases, I disagree with that statement if the view size is dependent on the text size. But it's great that you submitted that alternative fix. Thank you!

philipbrito: Glad to see that discussion here guys. @HusseinElFeky A detailed explanation 👏 .

hfeky: Glad to help! :smiley:

PR #2345, opened 2019-05-27
[FIX] SAML: Save username on login

Exordian:
@RocketChat/android

Detailed description:
When using SAML login, the `username` is not saved into the local repository, which is a requirement for e.g. sendMessage [1] [2] [3] (a nullpointer occurs on [3])

Fixes message send (and probably more) when using SAML login

[1] https://github.com/RocketChat/Rocket.Chat.Android/blob/1fd253cc9b8a92ff0d4b25b1d714653d113e9787/app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomPresenter.kt#L360
[2] https://github.com/RocketChat/Rocket.Chat.Android/blob/391a93618f16208d84418594e72f9c122fed6676/app/src/main/java/chat/rocket/android/infrastructure/LocalRepository.kt#L38
[3] https://github.com/RocketChat/Rocket.Chat.Android/blob/1fd253cc9b8a92ff0d4b25b1d714653d113e9787/app/src/main/java/chat/rocket/android/chatroom/presentation/ChatRoomPresenter.kt#L370


philipbrito (approved):
Thanks for sending this PR @Exordian!

PR #2350, opened 2019-05-29
Fix emoji popup (DE only)

Exordian:
@RocketChat/android

Fixes crash on emoji popup due to broken string format string


Exordian:
already resolved on master

PR #2353, opened 2019-05-31
[FIX] Channel Freeze Issue

Shailesh351:
@RocketChat/android

#### Changes:

- Fixed the latest channel freeze issue which shows black white screen and restarts app because of heavy uses of memory in background.
- This issue happens when getting mention string when both name and username is null

### DIFF
--- a/app/src/main/java/chat/rocket/android/helper/MessageParser.kt
+++ b/app/src/main/java/chat/rocket/android/helper/MessageParser.kt
@@ -70,8 +70,9 @@ class MessageParser @Inject constructor(
 
         message.mentions?.forEach {
             val mention = getMention(it)
-            if(mention.isNotEmpty())
+            if (mention.isNotEmpty()) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/chat/rocket/android/helper/MessageParser.kt
```diff
@@ -70,7 +70,8 @@ class MessageParser @Inject constructor(
 
         message.mentions?.forEach {
             val mention = getMention(it)
-            mentions.add(mention)
+            if(mention.isNotEmpty())
```
hfeky: There is a missing space here after `if`. It would also be better if you can add curly brackets for that if statement. :smiley:

hfeky (changes requested):

philipbrito (approved):
Thanks for sending this PR @Shailesh351! 🎉

PR #2429, opened 2019-07-25
[IMPROVEMENT] Add loading bar in Settings

### LINKED ISSUE #2357: [IMPROVEMENT] Long data loading on Settings view/ Settings view not ready when opening it
We can see that the Settings view is not ready (some data is missing) when opening it. After some time the data is being show to the user.

There is a delay when getting the:

- User avatar;
- User username;
- App version;
- Server version;
- Switch for send analytics option (when it is enabled);
- Delete account text.

We should make this screen/view ready when opening it or at least show a progress until we get all needed information/data.

weijia-yu:
@RocketChat/android

Closes #2357 

#### Changes: 

Add loading bar if it is loading the data

loading

### DIFF
--- a/app/src/main/java/chat/rocket/android/settings/presentation/SettingsPresenter.kt
+++ b/app/src/main/java/chat/rocket/android/settings/presentation/SettingsPresenter.kt
@@ -67,6 +67,7 @@ class SettingsPresenter @Inject constructor(
     fun setupView() {
         launchUI(strategy) {
             try {
+                view.showLoading()
                 currentServer?.let {
                     val serverInfo = retryIO(description = "serverInfo", times = 5) {
                         rocketChatClientFactory.get(it).serverInfo()
@@ -84,6 +85,7 @@ class SettingsPresenter @Inject constructor(
                         )
                     }
                 }
+                view.hideLoading()
             } catch (exception: Exception) {
                 Timber.d(exception, "Error getting server info")
                 exception.message?.let {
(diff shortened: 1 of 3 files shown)


philipbrito:
@weijia-yu Would be great if we hide all the screen information (because we can still see the placeholders for some text) and show it when it is really ready to de displayed. Could you do that?

weijia-yu:
@filipedelimabrito you mean basically showing a blank screen before all info shows up? If in that case, what if the loading fails, would it show an error dialogue?

philipbrito:
@weijia-yu I mean doing the same we do on the chat list: It has nothing except the toolbar and navigation bar and our loading view. When the chat list is ready to be presented to the users we hide the loading and show that list. If something fails we will show a toast.

app/src/main/java/chat/rocket/android/settings/presentation/SettingsPresenter.kt
```diff
@@ -84,13 +85,15 @@ class SettingsPresenter @Inject constructor(
                         )
                     }
                 }
+                view.hideLoading()
```
philipbrito: Not needed to call `hideLoading()` two times here. You can call it on the `finally {...}` block just one time instead.

philipbrito (changes requested):

philipbrito (approved):
@weijia-yu I've fixed the requested changes here! Thanks for sending this PR!

